PR #3, opened 2026-03-28
Fix bidi surrogate handling, ctx.font caching, emoji correction, and cleanup

Cklaus1:
## Summary
- **Fix astral codepoint misclassification in bidi**: `charCodeAt` only sees surrogate halves — switched to `codePointAt` and extended `classifyChar` for supplementary RTL/AL ranges (U+10800–10FFF, U+1E800–1EDFF, U+1EE00–1EEFF)
- **Cache `ctx.font` assignment**: skip redundant browser font-string parsing when the font hasn't changed between calls
- **Fix emoji correction retry**: don't cache a zero correction when `document.body` was null — retry once body becomes available
- **Warn on non-px font sizes**: `parseFontSize` now warns once per font when falling back to 16px default
- **Unify duplicate grapheme segmenters**: layout.ts and measurement.ts shared identical `Intl.Segmenter` singletons — consolidated into one export
- **Guard `mergeUrlLikeRuns`**: skip already-absorbed segments to prevent double-processing edge case
- **In-place mutation for `carryTrailingForwardStickyAcrossCJKBoundary`**: eliminates 4 array copies (length is unchanged, input is not reused)
- **Add alignment comment on `countPreparedLines`**: documents dual-implementation contract with `walkPreparedLines`
- **Add bundler note on `measure-harfbuzz.ts`**: flags `require('fs')` for browser-targeting bundlers

## Test plan
- [x] `bun test` — 44/44 pass, 99 expect() calls
- [x] `bun run check` — typecheck + lint clean (0 warnings, 0 errors)
- [ ] Browser accuracy sweep (`bun run accuracy-check`) to verify no regressions

### DIFF
--- a/src/analysis.ts
+++ b/src/analysis.ts
@@ -362,7 +362,7 @@ function mergeUrlLikeRuns(segmentation: MergedSegmentation): MergedSegmentation 
   const starts = segmentation.starts.slice()
 
   for (let i = 0; i < segmentation.len; i++) {
-    if (kinds[i] !== 'text' || !isUrlLikeRunStart(segmentation, i)) continue
+    if (texts[i]!.length === 0 || kinds[i] !== 'text' || !isUrlLikeRunStart(segmentation, i)) continue
 
     let j = i + 1
     while (j < segmentation.len && kinds[j] !== 'space' && kinds[j] !== 'zero-width-break') {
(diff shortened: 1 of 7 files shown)


chenglou:
Thanks for the thorough pass! Most of this has since landed in other forms: bidi now decodes surrogate pairs against generated Unicode data, there's one shared grapheme segmenter, the CJK carry mutates in place, and the HarfBuzz file is gone. One piece still applies: a zero emoji correction gets cached while `document.body` is null. Happy to take that as a small separate PR with a test if you're up for it. Closing this bundle since it can't rebase anymore.

PR #4, opened 2026-03-28
Bump dev deps and allow TypeScript 6

threepointone:
Update devDependencies: harfbuzzjs -> ^0.10.1, oxlint -> ^1.57.0, oxlint-tsgolint -> ^0.18.1, and keep tsgolint unchanged. Widen peer dependency for typescript to "^5 || ^6" in package.json and update bun.lock to match these dependency changes.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -6,66 +6,66 @@
       "name": "pretext",
       "devDependencies": {
         "@types/bun": "latest",
-        "harfbuzzjs": "^0.10.0",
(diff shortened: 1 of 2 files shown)


threepointone:
why is typescript even a peer dependency btw?

chenglou:
Because ask Codex lol. Thanks! I moved it to devDep. Gonna publish a new release that supports both js and ts

PR #7, opened 2026-03-29
fix(demo): text deselection in demos by skipping unchanged textContent writes

bytaesu:
# Before

https://github.com/user-attachments/assets/f1cbc51d-b751-49ff-9ef2-51dae7f5392d

# After

https://github.com/user-attachments/assets/dc6227ba-cef1-4f0c-b01b-4aa75f04cf91


chenglou:
Thanks for spotting the issue! I've merged this to preserve your authorship. Then I'm using another fix (a much more elaborate, but correct & generalized one) to solve the root issue and avoid DOM reads in those paths =)

bytaesu:
> Thanks for spotting the issue! I've merged this to preserve your authorship. Then I'm using another fix (a much more elaborate, but correct & generalized one) to solve the root issue and avoid DOM reads in those paths =)

Awesome 😁

PR #10, opened 2026-03-29
perf: speed up isCJK

### LINKED ISSUE #2: speeding up isCJK
So I used claude to write me a few implementations of isCJK to speed it up.

2 big observations:

1. An easy win for non-cjk strings (i.e. most of them) is an early check:
```
if (c < 0x3000) return false;
```

2. `for ... of` is much slower than `for (let i= 0; i < s.length; i++)` although the for loop is more complicated as it has to manually process multi-char code-points.

Here are all implementation + benchmark code you can run: https://gist.github.com/kjk/bdbea9d90c3bb0454fbe26353c521bfd

Here are results for `EarlyFast` which does early exit and changes `for .. of` into regular for loop.

It's code: https://gist.github.com/kjk/bdbea9d90c3bb0454fbe26353c521bfd#file-is_cjk-js-L38

## Bun

  | Scenario | Original | Regex | Early | Bisect | Unrolled | EarlyFast | ForLoop |
  |---|---|---|---|---|---|---|---|
  | Single CJK | 11M | 22M | 12M | 12M | 16M | **79M** | 57M |
  | Single latin | 20M | 24M | 49M | 21M | 22M | 40M | **36M** |
  | CJK string | 12M | 18M | 17M | 12M | 14M | **42M** | 42M |
  | Latin string | 1.5M | 4.7M | 3.6M | 3.0M | 3.5M | **17M** | 6.5M |
  | Mixed string | 6.5M | 5.9M | 9.9M | 8.6M | 13M | **62M** | 34M |

  ## Node

  | Scenario | Original | Regex | Early | Bisect | Unrolled | EarlyFast | ForLoop |
  |---|---|---|---|---|---|---|---|
  | Single CJK | 85M | 54M | 51M | 44M | 56M | **112M** | 110M |
  | Single latin | 59M | 72M | 83M | 55M | 74M | **115M** | 103M |
  | CJK string | 79M | 55M | 60M | 55M | 64M | 90M | **129M** |
  | Latin string | 4.2M | **57M** | 4.3M | 2.8M | 4.5M | 12M | 8.3M |
  | Mixed string | 14M | 15M | 16M | 10M | 15M | **41M** | 38M |

no1x0:
## Summary

Speeds up `isCJK()` by replacing the current `for...of` scan with an indexed `charCodeAt()` loop, an early non-CJK fast path, and explicit surrogate-pair decoding for astral CJK.

## Why

This follows the performance discussion in `#2`.

The expected impact is on `prepare()`-side analysis work. `layout()` is unaffected because CJK classification is already cached in `SegmentMetrics`.

## Scope

- changes only `isCJK()` in `src/analysis.ts`
- preserves the existing Unicode ranges
- preserves astral CJK handling
- adds one tiny direct stability test

## Validation

- `bun test`
- local `isCJK()` microbenchmark
- local `profilePrepare()` spot check on representative Japanese, Chinese, and mixed app text

Closes #2

### DIFF
--- a/src/analysis.ts
+++ b/src/analysis.ts
@@ -103,23 +103,36 @@ function containsArabicScript(text: string): boolean {
 }
 
 export function isCJK(s: string): boolean {
-  for (const ch of s) {
-    const c = ch.codePointAt(0)!
-    if ((c >= 0x4E00 && c <= 0x9FFF) ||
-        (c >= 0x3400 && c <= 0x4DBF) ||
-        (c >= 0x20000 && c <= 0x2A6DF) ||
-        (c >= 0x2A700 && c <= 0x2B73F) ||
-        (c >= 0x2B740 && c <= 0x2B81F) ||
-        (c >= 0x2B820 && c <= 0x2CEAF) ||
-        (c >= 0x2CEB0 && c <= 0x2EBEF) ||
-        (c >= 0x30000 && c <= 0x3134F) ||
-        (c >= 0xF900 && c <= 0xFAFF) ||
-        (c >= 0x2F800 && c <= 0x2FA1F) ||
-        (c >= 0x3000 && c <= 0x303F) ||
-        (c >= 0x3040 && c <= 0x309F) ||
-        (c >= 0x30A0 && c <= 0x30FF) ||
-        (c >= 0xAC00 && c <= 0xD7AF) ||
-        (c >= 0xFF00 && c <= 0xFFEF)) {
+  for (let i = 0; i < s.length; i++) {
+    const first = s.charCodeAt(i)
+    if (first < 0x3000) continue
+
+    if (first >= 0xD800 && first <= 0xDBFF && i + 1 < s.length) {
+      const second = s.charCodeAt(i + 1)
+      if (second >= 0xDC00 && second <= 0xDFFF) {
+        const codePoint = ((first - 0xD800) << 10) + (second - 0xDC00) + 0x10000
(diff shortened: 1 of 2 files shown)


anematode:
Presumably `isCJK` is usually false? e.g. you could have an early scan

```
let m = 0;
for (let i = 0; i < s.length; i++) {
    m |= s.charCodeAt(i);
}
if (m < 0x3000) return false;
// ... rest of your code ...
```

then most non-CJK scripts would pass quickly.

But also the existing uses of `isCJK` seem quite suboptimal. e.g. at line 763 of `analysis.ts`. it's checking isCJK on every array element twice

no1x0:
Thanks — I took a closer look at both parts of this.

The OR-scan idea seems valid as a negative-case prefilter, but it also adds a guaranteed extra pass for positive/mixed CJK cases, so I did not want to stack that onto this PR without stronger end-to-end prepare() evidence.

The call-site reuse point in src/analysis.ts looks more compelling, especially around the repeated adjacent/merged-segment checks. If I follow this up, I think that is more likely to be a separate small analysis-side change than another isCJK() tweak in this PR.

zori150:
hi

zori150:
hello

chenglou:
landed this on main in 2e4e4cb. i kept the same basic shape here: indexed UTF-16 scan, early < 0x3000 bail, and surrogate decoding only when needed. i also folded in the newer CJK extension coverage and added small durable tests. credited you in the commit trailer. closing this as superseded by the landed version. thanks.

PR #13, opened 2026-03-29
Use binary search for chunk lookup in layoutNextLine()

defpis:
## Summary

- Replace O(n) linear scan in `findChunkIndexForStart()` with O(n log n) binary search
- Fixes quadratic scaling when `layoutNextLine()` iterates through large `pre-wrap` texts with many hard breaks (`\n`)
- No behavioral change — only the lookup strategy is different

## Motivation

`layoutNextLine()` is the streaming line-by-line API, called repeatedly with advancing cursors. In `pre-wrap` mode each `\n` creates a chunk, and every call triggered a linear scan over all chunks to find the one containing the current segment index.

For a 5,000-line pre-wrap text this caused ~12.5M comparisons total (O(n²/2)), while the chunks array is sorted by `consumedEndSegmentIndex`, making binary search trivial.

## Benchmark

| Lines | Chunks | Linear (before) | Binary (after) | Speedup |
|-------|--------|-----------------|----------------|---------|
| 100   | 100    | 0.10ms          | 0.10ms         | —       |
| 1,000 | 1,000  | 2.00ms          | 0.80ms         | 2.5×    |
| 5,000 | 5,000  | 67.00ms         | 3.20ms         | 21×     |

`walkLineRanges()` and `layoutWithLines()` (which iterate chunks directly) are unaffected, confirming the bottleneck was in chunk lookup.

`normal` white-space mode has only 1 chunk, so the change is a no-op there.

## Test plan

- [x] `bun test` — 60/60 pass
- [x] `bun run check` — typecheck + lint clean
- [x] `bun run benchmark-check` — no regression on existing benchmarks

### DIFF
--- a/src/line-break.ts
+++ b/src/line-break.ts
@@ -92,11 +92,18 @@ function fitSoftHyphenBreak(
 }
 
 function findChunkIndexForStart(prepared: PreparedLineBreakData, segmentIndex: number): number {
-  for (let i = 0; i < prepared.chunks.length; i++) {
-    const chunk = prepared.chunks[i]!
-    if (segmentIndex < chunk.consumedEndSegmentIndex) return i
+  const chunks = prepared.chunks
+  let lo = 0
+  let hi = chunks.length - 1
+  while (lo <= hi) {
+    const mid = (lo + hi) >>> 1
+    if (segmentIndex < chunks[mid]!.consumedEndSegmentIndex) {
+      hi = mid - 1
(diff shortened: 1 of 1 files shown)


chenglou:
Thank you! I've made a small superset of this PR, so I'll close this one, but you're co-author on that commit so it preserves your contribution =). We also now have richer benchmarks for `layoutNextLine` now!

PR #14, opened 2026-03-29
perf: Replace Array.from with new Array in measureAnalysis

SamiKamal:
Replaces `Array.from<number>({ length: N })` with `new Array<number>(N)` for two temporary arrays in `measureAnalysis()`. Both are immediately populated by index assignment, so the `Array.from` iterator overhead is pure waste.

`Array.from({ length: N })` loops N times to fill the array with `undefined`, then we overwrite every slot anyway. `new Array(N)` just creates the array directly — no unnecessary loop.

### Benchmark

- https://jsperf.app/xidefu
- https://jsperf.app/xidefu/2

### DIFF
--- a/src/layout.ts
+++ b/src/layout.ts
@@ -214,8 +214,8 @@ function measureAnalysis(
   const breakableWidths: (number[] | null)[] = []
   const breakablePrefixWidths: (number[] | null)[] = []
   const segments = includeSegments ? [] as string[] : null
-  const preparedStartByAnalysisIndex = Array.from<number>({ length: analysis.len })
-  const preparedEndByAnalysisIndex = Array.from<number>({ length: analysis.len })
+  const preparedStartByAnalysisIndex = new Array<number>(analysis.len)
+  const preparedEndByAnalysisIndex = new Array<number>(analysis.len)
 
   function pushMeasuredSegment(
     text: string,


spaceemotion:
just fyi; switching to the regular `new Array(length)` will create a holey Array, and may not have the positive performance impact you'd think it'll have. Your jsperf link only checks the array instantiation, but not its usage.

See https://v8.dev/blog/elements-kinds#avoid-creating-holes

SamiKamal:
> just fyi; switching to the regular `new Array(length)` will create a holey Array, and may not have the positive performance impact you'd think it'll have. Your jsperf link only checks the array instantiation, but not its usage.
> 
> See https://v8.dev/blog/elements-kinds#avoid-creating-holes

It does create a holey array, but it gets filled [below](https://github.com/SamiKamal/pretext/blob/5614dd43c9a607a63d2f4496d4191ffc1187db9d/src/layout.ts#L244)

that being said, it won't make a HUGE impact, but it's a small free win

spaceemotion:
Have you tested the full function in context with the provided benchmarking tools? From a few personal spot checks, the change tended to make it worse (tested with Node, not Bun) - as holey arrays stay that way, and every read access will continue to be slower than a packed array.

SamiKamal:
I checked the actual read pattern. These arrays are only read in `mapAnalysisChunksToPreparedChunks`, which loops over `chunks.length` (1 in normal mode, one per \n in pre-wrap). 

Each iteration reads at most 3 slots from each array, so we're talking 3-6 total reads in the common case. The hole-check penalty on that few reads is negligible.

That said, after running the benchmarks  provided in this project, the allocation difference for arrays of size 10-120 is also negligible, so this change probably doesn't matter either way. Happy to close this PR.

SamiKamal:
Closing .. the performance gain is negligible and is mostly noise

PR #15, opened 2026-03-29
perf: Fast-path splitSegmentByBreakKind for pure-text segments

SamiKamal:
### What

`splitSegmentByBreakKind` is called for every segment produced by `Intl.Segmenter` during `prepare()`. It iterates each character with `for...of`, classifies it via `classifySegmentBreakChar`, and groups consecutive characters of the same kind into pieces.

For the vast majority of segments — normal text words like `"hello"`, `"world"`, `"المكتبة"` — every character classifies as `text`, producing a single piece. The per-character loop, string concatenation (`currentText += ch`), and array/object allocation for the pieces list are all unnecessary work in this case.

This PR adds a hoisted regex that matches the 9 characters `classifySegmentBreakChar` maps to non-`text` kinds (space, tab, newline, NBSP, NNBSP, word joiner, BOM, ZWSP, soft hyphen). If the segment doesn't contain any of them, we return a single-element array immediately, skipping the loop entirely.

### Benchmark

Using the project's built-in benchmark page (`/benchmark`), the analysis phase for Arabic text drops from ~29ms to ~12ms, a ~58% improvement.

### DIFF
--- a/src/analysis.ts
+++ b/src/analysis.ts
@@ -333,12 +333,18 @@ function classifySegmentBreakChar(ch: string, whiteSpaceProfile: WhiteSpaceProfi
   return 'text'
 }
 
+// All characters that classifySegmentBreakChar maps to a non-'text' kind.
+const breakCharRe = /[\x20\t\n\xA0\xAD\u200B\u202F\u2060\uFEFF]/
+
 function splitSegmentByBreakKind(
   segment: string,
   isWordLike: boolean,
   start: number,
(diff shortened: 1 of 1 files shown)


chenglou:
landed this on main in 2e4e4cb. i kept the narrow fast path here instead of the bigger rewrite variants we tried: cheap early return for pure-text segments, plus a tiny current-head measureAnalysis cleanup while i was in there. credited you in the commit trailer. closing this as superseded by the landed version. thanks.

PR #20, opened 2026-03-29
Fix lineCount inconsistency between layout() and layoutWithLines()

### LINKED ISSUE #11: layout() and layoutWithLines() return different lineCount for trailing collapsible spaces
## Summary

`layout()` (via `countPreparedLinesSimple`) and `layoutWithLines()` (via `walkPreparedLinesSimple`) can return different `lineCount` values when trailing collapsible whitespace overflows the line edge.

## Root Cause

In `line-break.ts`, the two code paths handle collapsible space overflow differently:

**`countPreparedLinesSimple`** (used by `layout()`):
```typescript
if (newW > maxWidth + lineFitEpsilon) {
  if (isSimpleCollapsibleSpace(kind)) continue  // ← skip (CSS-correct: space hangs)
  ...
}
```

**`walkPreparedLinesSimple`** (used by `layoutWithLines()`):
```typescript
if (newW > maxWidth + lineFitEpsilon) {
  if (canBreakAfter(kind)) {                    // ← canBreakAfter includes 'space'
    appendWholeSegment(i, w)                     //   so collapsible space triggers a break
    emitCurrentLine(i + 1, 0, lineW - w)
    ...
  }
}
```

The walk path is missing the `isSimpleCollapsibleSpace` guard that the count path has.

## Reproduction

Any text where `textWidth + spaceWidth > maxWidth`:

```typescript
const p = prepareWithSegments("word   ", "16px Arial")  // text + trailing spaces

const a = layout(p, maxWidth, lineHeight)                // uses count path
const b = layoutWithLines(p, maxWidth, lineHeight)       // uses walk path

// a.lineCount !== b.lineCount when trailing spaces overflow
```

For a concrete example: if "word" measures 40px and each space measures 5px, at `maxWidth = 42`:
- `layout()` → 1 line (spaces hang past edge, CSS-correct)
- `layoutWithLines()` → 2 lines (space triggers break, remaining spaces become line 2)

## Suggested Fix

Add the same `isSimpleCollapsibleSpace` check to `walkPreparedLinesSimple`, before the `canBreakAfter` check:

```typescript
if (newW > maxWidth + lineFitEpsilon) {
  if (isSimpleCollapsibleSpace(kind)) {  // ← add this
    i++
    continue
  }
  if (canBreakAfter(kind)) {
    ...
```

This matches CSS `white-space: normal` behavior: trailing collapsible whitespace hangs past the line edge without triggering a line break.

## How This Was Found

While porting the layout engine to Nim/WASM, we wrote tests asserting `countPreparedLines` and `walkPreparedLines` produce identical `lineCount` for all inputs. The trailing-space case was the only divergence.

yuribodo:
## Summary

- `walkPreparedLinesSimple` was missing the `isSimpleCollapsibleSpace` guard that `countPreparedLinesSimple` already has
- When a trailing collapsible space overflowed the line edge, the walk path treated it as a line break (via `canBreakAfter`) instead of letting it hang past the edge per CSS `white-space: normal` semantics
- This caused `layoutWithLines()` to report a higher `lineCount` than `layout()` for inputs where content after the hanging space would still fit on the same line
- Added a regression test with synthetic segments that reproduces the exact divergence

Fixes #11

## Test plan

- [x] `bun run check` passes (tsc + oxlint)
- [x] `bun test` passes (61/61, was 60)
- [x] New test fails without the fix, passes with it
- [x] Existing trailing-whitespace test still passes

### DIFF
--- a/src/layout.test.ts
+++ b/src/layout.test.ts
@@ -416,6 +416,35 @@ describe('layout invariants', () => {
     }])
   })
 
+  test('trailing collapsible spaces hang consistently across layout APIs', () => {
+    // Regression: countPreparedLines and walkPreparedLines must agree on
+    // lineCount when trailing collapsible whitespace overflows the line edge.
+    // Construct synthetic segments: [text:40px] [space:5px] [text:3px]
+    // At maxWidth=42: text(40) fits, text+space(45) overflows, but the space
+    // should hang (CSS-correct) and the 3px segment should stay on line 1.
+    // Without the fix, the walk path breaks at the space and puts the 3px
+    // segment on line 2.
+    const prepared = {
+      widths: [40, 5, 3],
+      lineEndFitAdvances: [40, 5, 3],
+      lineEndPaintAdvances: [40, 5, 3],
+      kinds: ['text' as const, 'space' as const, 'text' as const],
+      simpleLineWalkFastPath: true,
+      breakableWidths: [null, null, null],
+      breakablePrefixWidths: [null, null, null],
+      discretionaryHyphenWidth: 0,
+      tabStopAdvance: 0,
+      chunks: [{ startSegmentIndex: 0, endSegmentIndex: 3, consumedEndSegmentIndex: 3 }],
+    }
+
+    // maxWidth=44: text(40) fits, text+space(45) overflows, text+text2(43) fits
+    const countResult = countPreparedLines(prepared, 44)
+    const walkResult = walkPreparedLines(prepared, 44)
+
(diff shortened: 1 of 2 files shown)


chenglou:
Thanks for the careful repro! This got fixed the other way around. `layout()` and `layoutWithLines()` now share one line walker, so they can't disagree anymore. A trailing space can hang past the edge, but the word after it still has to fit, so in the test's [40, 5, 3] at 44px wrapping is correct. Closing as fixed on main.

PR #24, opened 2026-03-30
corpus(ko): replace Korean corpus with Sonagi by Hwang Sun-won

qtaghdi:
## Summary

- Replaces the Korean long-form stress corpus from Hyun Jin-geon's *Usu Joheun Nal* (`ko-unsu-joh-eun-nal`) with Hwang Sun-won's *Sonagi* (`ko-sonagi`)
- Updates all references: corpus page, benchmark snapshots, font-matrix script, `sources.json`, `README.md`, `STATUS.md`
- Reformats source text to remove fixed-width line wrapping; updates character/line counts in `sources.json`

## Test plan

- [ ] `bun run corpus-check --id=ko-sonagi 300 600 800` passes
- [ ] `/corpus` page loads and displays the new Korean corpus correctly
- [ ] `/benchmark` page includes `ko-sonagi` without errors


chenglou:
Thanks! Is this better as a replacement, or as an addition alongside `ko-unsu-joh-eun-nal`? Genuinely asking, since I'm unfamiliar with these corpuses. If `ko-sonagi` is a superset in terms of coverage, then yeah replacing the current one makes sense. Otherwise, we can also have both. I'm a bit torn. What you do think

chenglou:
Thanks! I've added this as a new Korean corpus in 08a5ba0 and credited your contribution

qtaghdi:
@chenglou Sorry for the late reply!

While "Unsu Joheun Nal" is a great work, I noticed that the version used in the Korean corpus had many line breaks that don’t align well with natural Korean formatting.

I tried to find a reliable original source to correct this, but unfortunately couldn’t find a clean version to work from.

So I chose "Sonagi", which is one of the most well-known Korean novels, and carefully reviewed and adjusted the line breaks myself. I believe this should result in more accurate measurements.

Thanks again for adding it!

Issue #28, opened 2026-03-30
Error message "Intl.Segmenter is not a constructor"


MichaelBrazier:
In my browser, all attempts to view the demos fail with an error message "Intl.Segmenter is not a constructor". As I can't find any class Intl.Segmenter in the Javascript that reaches my browser, it isn't surprising that the browser can't find a constructor for that class. But I'm wondering how the same code manages to work in other browsers.

xiaoiver:
Maybe you can try [grapheme-splitter](https://github.com/orling/grapheme-splitter) in non-browser environment.
https://infinitecanvas.cc/guide/lesson-015#cluster

For example in CI Node.js environment:
https://github.com/xiaoiver/infinite-canvas-tutorial/blob/master/__tests__/utils.ts#L149-L152

```ts
import GraphemeSplitter from 'grapheme-splitter';

const splitter = new GraphemeSplitter();
return splitter.splitGraphemes(s);
```

chenglou:
Documented the runtime requirement in cca3f1d: the main package needs Intl.Segmenter and Canvas 2D text measurement, so older browsers/runtimes without Intl.Segmenter are unsupported today. Thanks for flagging the confusing failure mode.

PR #30, opened 2026-03-30
docs: add browser requirements section to README

Luc0-0:
Fixes a gap that's been causing confusion — there's currently no mention of which browsers are supported, and issue #28 shows users hitting `Intl.Segmenter is not a constructor` with no guidance on why.

Adds a short **Browser Requirements** section before Caveats listing the minimum versions for `Intl.Segmenter` and Canvas `measureText`, and notes that older Firefox (pre-125) will throw on `prepare()`.


riquinhodo1171-star:
[Uploading index (1).html…]()

chenglou:
Thanks! I folded the runtime-requirements note into cca3f1d and credited you as co-author.

PR #32, opened 2026-03-30
added server-side canvas support & Created Masonry Demo

Unmesh100:
Masonry Demo :
- pages/demos/masonry.ts — Logic that computes card heights with Pretext.layout() without any DOM reads, then routes cards to the shortest column
- pages/demos/masonry.html — Dark-themed aesthetic with image-color cards, fixed bottom controls for column adjustment
- Updated pages/demos/index.html — Enabled the previously-linked but missing demo
Why this matters:
- Masonry layouts typically require expensive DOM measurement (getBoundingClientRect) to know heights before placement
- This demo routes cards to the shortest column using heights computed purely from canvas measurement — zero DOM reads in the hot path
- The column slider demonstrates dynamic reflow without any reflow penalty
Tech highlights:
- Uses layout(prepared, maxWidth, lineHeight) for height prediction
- Cards have variable heights based on their text content
- Column assignment via greedy shortest-column algorithm
- Responsive column count auto-adjusts to viewport


chenglou:
Thanks for this! There's already a masonry demo in pages/demos/masonry, so I'd rather not add a second one. Server-side measurement is still an open question for me. Server canvases don't measure like browsers do, and the line breaking is tuned against real browser widths. If you want to try it anyway, defining a global `OffscreenCanvas` backed by your server canvas works without new API. Closing this one, but thanks for building it!

PR #33, opened 2026-03-30
fix: add CJK Extension H/I to isCJK and fix pre-wrap fast path

xr843:
## Summary

Two small, targeted fixes:

1. **Add missing CJK Extension H and I ranges to `isCJK()`**

   CJK Unified Ideographs Extension H (U+31350–U+323AF, Unicode 15.0) and Extension I (U+2EBF0–U+2F7FF, Unicode 15.1) are not covered by the current range checks. Characters from these blocks fall through without CJK per-character line breaking or kinsoku attachment rules.

   This directly addresses the documented requirement in CLAUDE.md:
   > "Astral CJK ideographs, compatibility ideographs, and the later extension blocks must still hit the CJK path"

   The new ranges are placed adjacent to their neighboring extensions (I after F, H after G) to keep the block ordering clear.

2. **Remove no-op `.replace()` on `normalizeWhitespacePreWrap` fast path**

   When the guard `!/[\r\f]/.test(text)` is true, no `\r` exists in the text, so the subsequent `.replace(/\r\n/g, '\n')` can never match — it always returns the input unchanged. The fast path now returns `text` directly.

## Changes

- `src/analysis.ts`: Add Extension H/I ranges to `isCJK()`; simplify pre-wrap fast path
- `src/layout.test.ts`: Add Extension H/I ranges to `isWideCharacter()` test helper; add test case verifying Extension H/I characters get CJK break behavior

## Test plan

- [x] `bun test` — 61 tests pass (including new Extension H/I test)
- [x] `tsc --noEmit` — no type errors
- [x] Verified Extension H (U+31350) and Extension I (U+2EBF0) now return `true` from `isCJK()`
- [x] Verified the pre-wrap fast path is functionally equivalent (no `\r` → no `\r\n` → replace was always a no-op)

### DIFF
--- a/src/analysis.ts
+++ b/src/analysis.ts
@@ -112,7 +112,7 @@ export function isCJK(s: string): boolean {
         (c >= 0x2B740 && c <= 0x2B81F) ||
         (c >= 0x2B820 && c <= 0x2CEAF) ||
         (c >= 0x2CEB0 && c <= 0x2EBEF) ||
-        (c >= 0x2EBF0 && c <= 0x2F7FF) ||
+        (c >= 0x2EBF0 && c <= 0x2EE5F) ||
         (c >= 0x30000 && c <= 0x3134F) ||
         (c >= 0x31350 && c <= 0x323AF) ||
         (c >= 0xF900 && c <= 0xFAFF) ||
--- a/src/layout.test.ts
+++ b/src/layout.test.ts
@@ -42,7 +42,7 @@ function isWideCharacter(ch: string): boolean {
     (code >= 0x2B740 && code <= 0x2B81F) ||
     (code >= 0x2B820 && code <= 0x2CEAF) ||
     (code >= 0x2CEB0 && code <= 0x2EBEF) ||
-    (code >= 0x2EBF0 && code <= 0x2F7FF) ||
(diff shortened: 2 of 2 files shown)


xr843:
Thanks for the careful review @abnow-dev! Both issues are now fixed:

1. **Extension H upper bound**: corrected from `0x2F7FF` to `0x2EE5F`
2. **Swapped H/I labels**: fixed the code points and Unicode version labels in the test comments — Extension H is `0x2EBF0` (Unicode 15.1), Extension I is `0x31350` (Unicode 16.0)

All tests pass.

chenglou:
Thanks for the careful catch here, Tim. The core CJK extension coverage this PR was targeting has since landed on  as part of a broader follow-up (), including the matching test/helper updates, so I’m going to close this as superseded rather than leave an outdated PR hanging around. Really appreciate you surfacing it and iterating on the extension bounds.

chenglou:
Small follow-up for clarity: current main already includes the CJK extension coverage this PR was aiming for, via the later upstream change in 2e4e4cb, along with the matching test/helper coverage. Thanks again for surfacing it and iterating on the bounds.

Issue #1, opened 2026-03-28
Request: vertical-rtl


aehlke:
Thank you for the incredible piece of work. It is revolutionary for many needs.

From inspecting the code it looks like this doesn't support vertical text. Could you please add vertical-rtl (Japanese)? Or give us guidance on how to add it? Much appreciated.

tallman1O:
@aehlke i have opened PR addressing this issue, please check it out

chenglou:
Sorry for coming back to this only now! 'm open to adding vertical support, but only if there are potentially prominent use-cases (as this'd expand the api and coverage quite a lot)

tallman1O:
hi @chenglou i love reading japanese manags and light novels, mostly in english but i am shifting to native japanese language, i have noticed that they still default to vertical layout so this would be really beneficial for publishing these books digitally and any reading app or CMS targeting the Japanese market pretty much needs it. i think css also supports this vertical property. this will allow pretext for an entire category of e-reader/publishing tools

aehlke:
> hi [@chenglou](https://github.com/chenglou) i love reading japanese manags and light novels, mostly in english but i am shifting to native japanese language, i have noticed that they still default to vertical layout so this would be really beneficial for publishing these books digitally and any reading app or CMS targeting the Japanese market pretty much needs it. i think css also supports this vertical property. this will allow pretext for an entire category of e-reader/publishing tools

I maintain a popular Japanese ebook reading app that I'd like to use this in. But ebooks and readability mode web content often need CSS styling that this library doesn't yet support.

PR #37, opened 2026-03-30
Add vertical Japanese layout support

tallman1O:
Introduced a new vertical Japanese layout option (`vertical-rl`) in the API, allowing for top-to-bottom text flow with right-to-left column stacking. Updated the README with usage instructions and examples. Enhanced type definitions to include `WritingMode` and added tests to validate the new layout behavior.

Fixes Issue : Request: vertical-rtl #1


ghost:
:)

Issue #50, opened 2026-03-30
Bug: layoutNextLine() produces different line breaks than layoutWithLines() on mixed-script text (CJK + RTL + emoji)


SeaL773:
## Summary

`layoutNextLine()` and `layoutWithLines()` produce different line breaks for the same prepared text and width when the input contains mixed scripts (Latin + CJK + Arabic RTL + emoji).

This appears to be a separate trigger from #43 (ZWSP + space leading space behavior). Here the inconsistency is in how CJK segment boundaries interact with the line-start normalization logic between the batch and iterator APIs.

## Reproduction

```ts
import { prepareWithSegments, layoutWithLines, layoutNextLine } from '@chenglou/pretext'

const text = 'Hello \u4e16\u754c \u0645\u0631\u062d\u0628\u0627 \ud83c\udf0d test'
const font = '16px Inter'
const width = 80
const lineHeight = 19

const prepared = prepareWithSegments(text, font)

// Batch API
const batch = layoutWithLines(prepared, width, lineHeight)
console.log(batch.lineCount) // 3
console.log(batch.lines.map(l => l.text))
// Line 0: 'Hello '  (start=seg0, end=seg2)
// Line 1: '\u754c \u0645\u0631\u062d\u0628\u0627 '  (start=seg3, end=seg7)
// Line 2: '\ud83c\udf0d test'  (start=seg7, end=seg10)
// Note: '\u4e16' (seg2) is missing from the output

// Iterator API
const lines = []
let cursor = { segmentIndex: 0, graphemeIndex: 0 }
while (true) {
  const line = layoutNextLine(prepared, cursor, width)
  if (line === null) break
  lines.push(line)
  cursor = line.end
}
console.log(lines.length) // 4
console.log(lines.map(l => l.text))
// Line 0: 'Hello '
// Line 1: '\u4e16\u754c '
// Line 2: '\u0645\u0631\u062d\u0628\u0627 \ud83c\udf0d '
// Line 3: 'test'
```

## Key observation

`layoutWithLines` produces 3 lines where line 1 starts at segment index 3 -- it skipped segment 2. The character at segment 2 appears to be consumed by the end cursor of line 0 but not rendered in line 0's text. `layoutNextLine` correctly handles this and produces 4 lines.

Segment breakdown:
```
Index: 0       1    2       3       4    5          6    7       8    9
Text:  Hello   ' '  \u4e16  \u754c  ' '  \u0645..  ' '  \ud83c  ' '  test
[...]

PR #52, opened 2026-03-31
fix: unify countPreparedLines with walkPreparedLines to fix lineCount mismatch (#49)

### LINKED ISSUE #49: Bug: layout() and layoutWithLines() disagree on lineCount when ZWSP forces overflow-wrap grapheme breaking at narrow widths
## Summary

`layout()` (via `countPreparedLinesSimple`) and `layoutWithLines()` (via `walkPreparedLinesSimple`) return different `lineCount` values when text contains a ZWSP (`U+200B`) break opportunity and the layout width is narrow enough to trigger overflow-wrap grapheme-level breaking.

This is a separate trigger from #11 (trailing collapsible spaces). Here the divergence occurs in the overflow-wrap grapheme-breaking codepath, not the trailing-space handling.

## Reproduction

```ts
import { prepare, prepareWithSegments, layout, layoutWithLines } from '@chenglou/pretext'

const text = 'alpha\u200Bbeta'
const font = '16px Inter'
const lineHeight = 19
const maxWidth = 10 // narrow enough to force grapheme breaking

const p1 = prepare(text, font)
const p2 = prepareWithSegments(text, font)

const a = layout(p1, maxWidth, lineHeight)
const b = layoutWithLines(p2, maxWidth, lineHeight)

console.log(a.lineCount) // 10
console.log(b.lineCount) // 9  <-- disagrees!
```

Also reproduces with ZWSP + ZWNJ combination:

```ts
const text2 = 'alpha\u200Bbeta\u200Cgamma'
// layout() -> 15, layoutWithLines() -> 14
```

## Root Cause (preliminary)

`countPreparedLinesSimple` and `walkPreparedLinesSimple` handle the `placeOnFreshLine` / overflow-wrap grapheme-breaking path differently.

In `countPreparedLinesSimple`, `placeOnFreshLine` increments `lineCount` with a different counting strategy than `walkPreparedLinesSimple`'s equivalent `appendBreakableSegment` + `emitCurrentLine` sequence, leading to an off-by-one in the total line count when a ZWSP creates a break point that splits into grapheme-breakable segments at narrow widths.

## Environment

- Tested with bun 1.3.11 on Windows using the fake canvas backend from layout.test.ts
- All existing `bun test` suite passes (60/60)
- Found via systematic fuzz testing of API consistency invariants

SeaL773:
## Summary

Fixes #49.

`layout()` and `layoutWithLines()` returned different `lineCount` values when text contained ZWSP (`U+200B`) break opportunities at narrow widths that triggered overflow-wrap grapheme-level breaking.

## Root Cause

`countPreparedLinesSimple` (used by `layout()` via `countPreparedLines`) had two divergences from `walkPreparedLinesSimple` (used by `layoutWithLines()`):

1. **Grapheme-breaking counting**: The count path incremented `lineCount` at the **start** of each grapheme line (`if (lineW === 0) lineCount++`), while the walk path only emits lines at **overflow points** or at the end. This caused an off-by-one when a breakable segment's graphemes all fit within `maxWidth`.

2. **Missing `canBreakAfter` branch**: The count path's overflow handler unconditionally reset `hasContent` and called `placeOnFreshLine`, while the walk path first checked `canBreakAfter(kind)` to append the current segment to the line before emitting. This caused segments like ZWSP to be placed on their own line in the count path but merged into the previous line in the walk path.

## Fix

Remove the separate `countPreparedLinesSimple` fast path and always delegate `countPreparedLines` to `walkPreparedLines` (without an `onLine` callback). This guarantees `layout()` and `layoutWithLines()` always produce identical `lineCount`.

## Before / After

```ts
const text = 'alpha\u200Bbeta'  // ZWSP between words
const prepared = prepare(text, '16px Inter')

// Before fix (at maxWidth = 10):
layout(prepared, 10, 19).lineCount           // 10
layoutWithLines(prepared, 10, 19).lineCount  // 9  <-- disagrees!

// After fix:
layout(prepared, 10, 19).lineCount           // 9
layoutWithLines(prepared, 10, 19).lineCount  // 9  <-- agrees!
```

## Testing

- All 60 existing tests pass (`bun test src/layout.test.ts`)
- Verified with systematic fuzz testing (361 tests across 23 text variants x multiple widths): `layout()`, `layoutWithLines()`, `walkLineRanges()`, and `layoutNextLine()` now agree on lineCount for all tested inputs (except for a separate `layoutNextLine` issue tracked in #50)

## Performance Note

This change removes the count-only fast path, which means `layout()` now runs the full walk logic internally (without materializing line objects since no `onLine` callback is passed). The walk path already skips string construction and line-object allocation when `onLine` is undefined, so the practical overhead should be minimal. If performance regression is a concern, the count path could be restored by carefully porting the walk path's `canBreakAfter` branch and deferred grapheme counting — but given the subtlety of the divergence, I believe correctness should take priority.

### DIFF
--- a/src/line-break.ts
+++ b/src/line-break.ts
@@ -131,9 +131,11 @@ export function normalizeLineStart(
 }
 
 export function countPreparedLines(prepared: PreparedLineBreakData, maxWidth: number): number {
-  if (prepared.simpleLineWalkFastPath) {
-    return countPreparedLinesSimple(prepared, maxWidth)
-  }
+  // Unify with walkPreparedLines to guarantee layout() and layoutWithLines()
+  // always agree on lineCount. The previous dedicated countPreparedLinesSimple
+  // fast path diverged from walkPreparedLinesSimple in the overflow-wrap
+  // grapheme-breaking and canBreakAfter codepaths, causing off-by-one
+  // lineCount mismatches. See #49.
   return walkPreparedLines(prepared, maxWidth)
 }
 


chenglou:
Thanks for the writeup.

This is superseded on `main` by `f707bb2` now. We ended up taking the same lineCount-alignment fix in a slightly narrower form, with current regression coverage in `src/layout.test.ts`.

So I’m closing this PR rather than merging it as-is.

PR #61, opened 2026-03-31
bun start available in Windows

Hurrieam:
Supported by Doubao, bun start available in Windows, simplify start script and enable hot reload: Replace the long shell start script (which attempted to free/kill processes and read HOST/PORT env vars) with a simpler bun invocation using --hot and explicit --host=127.0.0.1 --port=3000. This removes the automatic port-killing logic and hardcodes the host/port for local development. Note: start:lan previously relied on setting HOST before running start, but the new start no longer reads HOST/PORT env vars—adjust scripts if you need LAN hosting or dynamic ports.


chenglou:

PR #65, opened 2026-03-31
Add LTR mark to README example string for bidi readability

cxa:
So confusing at first glance.
 
before:
Screenshot 2026-03-31 at 23 47 15

after:
Screenshot 2026-03-31 at 23 47 38


chenglou:
lol thanks!

PR #66, opened 2026-03-31
compatible-windows-start

xiaobai-web715:
The project failed to start on Windows. I replaced the startup on Windows using Bun.server, without modifying the demos files, only adding start.ts for startup configuration, while still keeping the original start command.


xiaobai-web715:
Use cross-env and rimraf to be compatible with Windows execution

xiaobai-web715:
image
Window startup routing output
image
demos/masonry router redirection causes relative path referenced resources to return 404（+ '/' ）
image
image
Adapt JSON/SVG/md resources

chenglou:

PR #67, opened 2026-03-31
add bun start for windows

moongazers:
bun start-windows


chenglou:

PR #69, opened 2026-03-31
Cleanup files to be published to npm

LitoMore:
It looks like some files are published to npm by accident.

What was the reasoning behind uploading those files to npm previously?

In the meantime, here is why I remove them:

- CHANGELOG.md - Useless for package distribution
- src/ - Useless for package distribution
- pages/ - Useless for package distribution
- LICENSE - npm included this file by default, you don't need to list the file here

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,14 +19,7 @@
     "./package.json": "./package.json"
   },
   "files": [
-    "CHANGELOG.md",
-    "dist",
-    "src",
-    "!src/layout.test.ts",
-    "!src/test-data.ts",
-    "pages/demos",
-    "pages/assets",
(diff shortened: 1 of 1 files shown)


chenglou:
Hey thanks! src and pages are intentional since they're still small, plus changelog. These have been proven useful for AI to explore repos that are not in the data distribution yet (such as this one), esp during upgrade. And this repo's api's so low level that the demos are important high-level patterns before we solidify more high-level apis

However the LICENSE thing is true so I've made the change. I've added you as co-author to preserve your authorship. Thanks!

LitoMore:
Got it. Thanks for your clarification.

PR #71, opened 2026-04-01
feat: add a compat entrypoint for portable regex support

Huxpro:
## Summary
- keep the default entrypoint on the original Unicode property regex literals
- add a new `@chenglou/pretext/compat` entrypoint for engines that cannot parse `\p{...}` literals
- cover the new subpath with tests and verify it matches the default layout behavior

## Why
Some runtimes fail while parsing Unicode property regex literals, before any feature detection can run. A separate compat entrypoint avoids changing the default browser bundle while still giving hosts like Lynx a portable option.

## Verification
- `bun x tsc -p tsconfig.build.json`


chenglou:
Thanks for this! A second copy of the engine for portable regex would drift with every engine change, and the runtime requirement is documented in the README now. Closing, thanks!

Issue #77, opened 2026-04-01
[BUG] Pretext bidi: canvas.measureText vs DOM width divergence


sepcnt:
Pretext exposes segLevels in PreparedTextWithSegments, enabling consumers to build bidi-aware coordinate maps. However, the underlying segment widths come from canvas.measureText(segment) on isolated segments. For Arabic script the browser's DOM renderer may produce different effective widths due to contextual shaping, directional spacing, or font-level bidi adjustments.

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<body link="#467886" vlink="#96607D">

Case | dir | bidi? | max Δ (px) | avg Δ (px)
-- | -- | -- | -- | --
LTR   baseline | ltr | no | 0.3 | 0.1
Short   Arabic (LTR paragraph) | ltr | yes | 7.4 | 4.1
Arabic   words (LTR paragraph) | ltr | yes | 8.4 | 4
Arabic   words (RTL paragraph) | rtl | yes | 8.4 | 4
Mixed   LTR+Arabic+LTR | ltr | yes | 78.6 | 45.9
Arabic +   numbers | ltr | yes | 8.4 | 1.8
LTR +   short Arabic | ltr | yes | 40.3 | 32.7

</body>

</html>

The PoC is attached below:

```html
<!DOCTYPE html>
<html>
<head>
<meta charset="utf-8">
<title>Pretext bidi: canvas vs DOM width divergence</title>
<style>
* { box-sizing: border-box; }
body { font: 15px/1.6 system-ui, sans-serif; max-width: 960px; margin: 2em auto; padding: 0 1em; }
h1 { font-size: 1.3em; }
h2 { font-size: 1.05em; margin-top: 2em; border-bottom: 1px solid #ddd; padding-bottom: .3em; }
table { border-collapse: collapse; margin: .8em 0; font-size: 13px; }
th, td { border: 1px solid #ddd; padding: 3px 8px; text-align: right; }
th { background: #f7f7f7; }
.bad { background: #fee; }
.ok  { background: #efe; }
pre { background: #f5f5f5; padding: .8em 1em; overflow-x: auto; font-size: 12px; border-radius: 4px; }
code { background: #f0f0f0; padding: .1em .3em; border-radius: 3px; font-size: .92em; }
[...]

PR #82, opened 2026-04-02
Fix bidi paragraph direction and add canvas-to-DOM shaping correction for RTL segments

sepcnt:
https://github.com/chenglou/pretext/issues/77
## TL;DR

Mixed left-to-right / right-to-left text (think English paragraphs with Arabic words in the middle) was always rendering wrong. Turns out the code that decides "is this paragraph LTR or RTL?" had a math bug that made *every* bidi paragraph RTL. On top of that, Arabic and Hebrew text was being measured slightly wrong because canvas and the browser's DOM disagree about how wide shaped glyphs are. Both issues are fixed, with zero impact on non-bidi text.

### The paragraph direction bug

When text contains both LTR and RTL characters, the bidi algorithm needs to pick a "base direction" for the paragraph. Our code tried to decide this with a ratio:

```typescript
const startLevel = (len / numBidi) < 0.3 ? 0 : 1
```

The intent was: "if bidi characters are a small fraction, call it LTR." But `len` is the total character count and `numBidi` is the bidi character count — so `len / numBidi` is always ≥ 1, which is always > 0.3, which means `startLevel` was always 1 (RTL).

A sentence like `"one اثنان three"` would be treated as an RTL paragraph with LTR islands instead of the correct LTR paragraph with an RTL island. The words came out in the wrong order.

### The width measurement gap

Canvas `measureText` and the browser's DOM layout engine don't agree on how wide Arabic and Hebrew text is. The DOM runs the full HarfBuzz shaping pipeline (cursive joining, ligatures, kerning), while canvas sometimes takes shortcuts. The difference is small per word, but it adds up over a line and can push line breaks to the wrong place.

### Paragraph direction: follow the spec

Replaced the broken ratio with the standard Unicode rule (UBA P2/P3): scan from the start of the paragraph and use the first "strong" character's direction.

```typescript
let startLevel = 0
for (let i = 0; i < len; i++) {
  const t = types[i]!
  if (t === 'L') { startLevel = 0; break }
  if (t === 'R' || t === 'AL') { startLevel = 1; break }
}
```

Simple, correct, and what every other bidi implementation does.

### Width correction: measure once, apply cheaply

`pretext` already had a similar trick for emoji (one cached DOM measurement per font). The RTL correction follows the same pattern:

1. **Sample** — pick up to 8 non-space RTL segments from the text (≤ 200 chars total), spread evenly so we get a representative slice.
2. **Measure** — concatenate them into one string, measure via canvas, measure via a hidden DOM `<span>`. One `getBoundingClientRect` call.
3. **Cache** — store the canvas-to-DOM ratio keyed by (font, sample text). Same RTL content in the same font hits cache next time.
4. **Apply** — multiply each RTL segment's canvas width by the ratio during `prepare()`.

A few things tried but rejected along the way:

- **Correcting all scripts, not just RTL** — this made Japanese line counts jump by 5.3%. CJK, Latin, and Thai don't have contextual shaping issues, so their canvas measurements are already accurate. Applying a correction to them just adds error.
- **Using hardcoded sample strings per script** — this would need maintenance every time we encounter a new font or script variant. Sampling from the actual text adapts automatically.

### Tests

13 new tests covering paragraph direction (LTR-first, RTL-first, short Arabic in LTR, numbers in RTL, pure LTR), visual ordering (mixed text, pure Arabic reversal, numbers in RTL context), structural invariants (matching array lengths, non-negative widths), and layout consistency (monotonic line counts, agreement between layout methods, round-trip text reconstruction).

## Performance

Benchmarked on a 500-text corpus in Chrome:

| Metric | Before | After | Δ |
|---|---|---|---|
| `prepare()` 500 texts | 19.75 ms | 19.10 ms | −3.3% |
| `layout()` 500 texts | 0.12 ms | 0.12 ms | — |
| Arabic long-form `prepare()` (cold) | 71.90 ms | 71.80 ms | — |
| Arabic long-form `layout()` (hot) | 0.24 ms | 0.23 ms | — |
| Arabic `layoutWithLines()` 8 texts | 6.17 ms | 6.64 ms | +7.6% |

Line counts for every non-Arabic script are identical to before (Japanese 380, Chinese 625, Korean 503, Thai 1013, etc.). Arabic prose is also unchanged at 2433 lines.

## Verification

- [x] All 84 tests pass (71 existing + 13 new bidi tests)
- [x] Benchmark shows no regression on the layout hot path
- [x] CJK / Latin / Thai line counts unchanged
- [x] Arabic line count unchanged
- [x] Verify page confirms correct visual ordering for mixed LTR+Arabic+LTR (previously broken)

### DIFF
--- a/src/layout.ts
+++ b/src/layout.ts
@@ -49,8 +49,8 @@ import {
 } from './analysis.js'
 import {
   clearMeasurementCaches,
-  getBidiCorrection,
   getCorrectedSegmentWidth,
+  getShapingRatio,
   getEngineProfile,
   getFontMeasurementState,
   getSegmentGraphemePrefixWidths,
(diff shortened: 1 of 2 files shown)


chenglou:
Thanks again for digging into this. I landed the narrow paragraph-direction fix separately in 4e24685 and kept your authorship on that commit with a Co-authored-by trailer, so the shipped fix still credits you.

That commit also adds permanent invariants around the rich bidi metadata shape: pure LTR still skips `segLevels`, and mixed bidi text now follows the first strong character for the paragraph base direction.

I’m going to close this PR as scoped because I don’t want to merge the wider shaping-ratio path yet. The `startLevel` bug was real and worth landing; the richer DOM/canvas correction still needs a tighter browser-backed oracle for the actual `#77` seam before we start rewriting rich-path widths. I’ll keep `#77` open for that larger question.

PR #105, opened 2026-04-05
fix: currency symbols stick to adjacent numbers during line breaking

### LINKED ISSUE #98: Wrapping of "$___" is inconsistent with browser
I noticed a case where Pretext's wrapping is inconsistent with browser behavior: https://codepen.io/editor/rf3521dcb4b0/pen/019d55dd-d8c4-750a-8b92-cd8d69641e0b

Image

This repro uses `prepareWithSegments`/`layoutWithLines`, although I originally saw it using the inline flow API.

ImpulseB23:
Fixes #98

Currency symbols (`$`, `£`, `€`, `¥`, etc.) were getting segmented separately from adjacent text by `Intl.Segmenter`, leaving a breakable gap. So pretext could wrap between `$` and `500`, or between `500` and `€`, which browsers don't do.

Added a `\p{Sc}` (Unicode currency symbol category) check to both the forward-sticky and left-sticky merge passes in `analysis.ts`. This covers prefix (`$500`) and postfix (`500€`) currencies, matching browser behavior.

Before:
- `$500` -> segments `["$", "500"]` (breakable)
- `500€` -> segments `["500", "€"]` (breakable)

After:
- `$500` -> segments `["$500"]`
- `500€` -> segments `["500€"]`

Tested in Chromium on Windows across `$`, `£`, `€`, `¥`, `₹` at various widths. All match browser line counts. 81/81 existing tests pass.

### DIFF
--- a/src/analysis.ts
+++ b/src/analysis.ts
@@ -97,6 +97,7 @@ export function setAnalysisLocale(locale?: string): void {
 
 const arabicScriptRe = /\p{Script=Arabic}/u
 const combiningMarkRe = /\p{M}/u
+const currencySymbolRe = /\p{Sc}/u
 const decimalDigitRe = /\p{Nd}/u
 
 function containsArabicScript(text: string): boolean {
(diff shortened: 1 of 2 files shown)


chenglou:
Thanks, this was a great catch. I landed a slightly broader version in 4d68871: instead of special-casing only currency symbols, it follows the numeric prefix/postfix line-break shape, so it covers `$___`, `$500`, `500€`, `+500`, `foo%bar`, `50°C`, etc. Credited you as co-author on the commit. Closing this PR since the fix is on main now. Appreciate it!

PR #108, opened 2026-04-05
feat: letterSpacing option in prepare()

### LINKED ISSUE #78: Feature request: letterSpacing support in prepare()
## Summary

Canvas `measureText()` does not account for `letter-spacing`, so text width calculations are inaccurate when letter-spacing is applied. Since pretext already segments text into graphemes, it seems like a natural place to handle this correction.

## Proposal

Add an optional `letterSpacing` parameter (in px) to `PrepareOptions`:

```ts
const prepared = prepare(text, font, { letterSpacing: 2 })
```

During `prepare()`, each segment's width would be adjusted by `graphemeCount × letterSpacing`. The `layout()` hot path would remain unchanged since it already operates on cached widths.

### Edge cases to consider

- CSS does not add spacing after the last character in a line
- Interaction with `overflow-wrap: break-word` grapheme-level breaking

## Why pretext is the right layer for this

Since canvas `measureText()` ignores letter-spacing entirely, any consumer of pretext that needs letter-spacing would have to reimplement segment-level grapheme counting — which pretext already does internally. Baking it in would keep measurement accurate and self-contained.

ImpulseB23:
Closes #78

Adds `letterSpacing` (in px) to `PrepareOptions`. During `prepare()`, each segment's cached width is adjusted by `graphemeCount * letterSpacing`, with trailing letter-spacing trimmed at line boundaries via `lineEndFitAdvance`. The `layout()` hot path is untouched.

Handles prefix currencies, CJK, pre-wrap, negative spacing (tighter tracking), and overflow-wrap grapheme-level breaking. Grapheme counting uses `Intl.Segmenter` and is gated behind `letterSpacing !== 0` so there's zero overhead on the default path.

```ts
const prepared = prepare(text, font, { letterSpacing: 2 })
const { height } = layout(prepared, width, lineHeight)
```

Tested in Chromium on Windows across various texts, widths, and spacing values (0-10px). All match browser line counts. 87/87 tests pass including 7 new letter-spacing tests.

### DIFF
--- a/src/layout.test.ts
+++ b/src/layout.test.ts
@@ -727,6 +727,61 @@ describe('rich-inline invariants', () => {
 })
 
 describe('layout invariants', () => {
+  test('letterSpacing increases segment widths by graphemeCount * spacing', () => {
+    const base = prepareWithSegments('Hello World', FONT)
+    const spaced = prepareWithSegments('Hello World', FONT, { letterSpacing: 2 })
+
+    // "Hello" = 5 graphemes, " " = 1 grapheme, "World" = 5 graphemes
+    expect(spaced.widths[0]).toBeCloseTo(base.widths[0]! + 5 * 2, 5)
+    expect(spaced.widths[1]).toBeCloseTo(base.widths[1]! + 1 * 2, 5)
+    expect(spaced.widths[2]).toBeCloseTo(base.widths[2]! + 5 * 2, 5)
+  })
+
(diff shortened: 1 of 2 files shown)


chenglou:
Thanks for the implementation sketch here. Numeric px `letterSpacing` support landed on `main` in 1a8b2ae, credited this PR!

PR #109, opened 2026-04-06
feat: resolve rem/em font units to px before canvas measurement

okaris:
## Problem

Canvas `ctx.font` silently ignores relative CSS units (`rem`, `em`). When callers pass CSS-derived font strings like `"0.9rem Inter"`, canvas keeps its previous font and returns wrong widths. This is a spec limitation — canvas can exist without a document root, so relative units have no reference point ([WHATWG HTML #1682](https://github.com/whatwg/html/issues/1682)).

In practice, anyone using Tailwind or a CSS framework with custom `text-sm` sizes (e.g. `--text-sm: 0.9rem` instead of the default `0.875rem`) gets silently wrong measurements. The text measures at whatever font was previously set on the canvas context, not at the intended size.

Related: #107 (canvas doesn't account for certain CSS properties).

## Solution

Add `resolveFont()` in `measurement.ts` that rewrites relative units to absolute px before the string reaches canvas:

```
"0.9rem Inter" → "14.4px Inter"  (given 16px root)
"bold 1.5em mono" → "bold 24px mono"
"14px Inter" → "14px Inter"  (no-op, no allocation)
```

Resolution uses `getComputedStyle(document.documentElement).fontSize` — the same level of DOM access that `getEmojiCorrection()` already does. Results are cached at module level so the DOM read happens once. Environments without a document (OffscreenCanvas, Node.js) fall back to 16px.

## Changes

- `resolveFont(font)` — exported, cached, handles rem/em → px
- `getFontMeasurementState()` — resolves font before `ctx.font =` and cache keying
- `parseFontSize()` — resolves before extracting the px value
- `clearMeasurementCaches()` — clears the resolution cache + cached root font size
- Tests for px passthrough, rem resolution, em resolution, and parseFontSize integration

## Test plan

- [x] `bun test src/layout.test.ts` — 85 tests pass (5 new)
- [x] Existing tests unaffected (all use px fonts, passthrough is zero-cost)
- [x] No document in test env → falls back to 16px root, rem math still correct


okaris:
Rebased onto current `main` and force-pushed — sorry for the churn on an old PR.

The previous version accidentally included built `dist/` output (16 files, +4343), which both buried the actual change and made the PR conflict with every release. That commit is dropped; `dist` is gitignored upstream and stays that way.

What is left is the original change only:

- `src/measurement.ts` (+46/-5) — resolve `rem`/`em` to `px` before assigning `ctx.font`, since canvas silently ignores relative units. Cached root font-size read, cached per-font-string result, absolute fonts pass through untouched with no allocation.
- `src/layout.test.ts` (+31) — coverage for rem resolution and `parseFontSize` on resolved values.

`measurement.ts` rebased with no conflicts. The only conflict was in the test file, where `main` had since added its own `MeasurementModule` type and `measurementMod` import — I kept yours and dropped my now-duplicate ones.

Verified on the rebased branch: `bun test` 121 pass / 0 fail, `tsc` clean.

Happy to split it further or adjust the caching approach if you would prefer something different.

PR #113, opened 2026-04-07
demo: add Knuth-Plass optimal line-breaking with fitness classification

ezra-en:
I've dreamed of a modern Knuth-Plass rendering solution since I discovered it, but all the independent implementations were last updated years ago. Since pretext provides both the foundation and the hype (completely admit jumping onto the bandwagon for this, but it really seems like noone's implemented knuth-plass yet) here's my own demo implementation!

I made this for my own typography-centric software but I thought it'd do more good upstreamed. Unsure if it should be added as another layout option, but I figured including it as a demo would probably suffice for devs using LLMs and looking for Knuth-Plass.


chenglou:
Thanks for building this! We've since shipped a justification comparison demo that already includes a Knuth-Plass-style layout, so this one's covered. Closing, but thanks again for the effort!

PR #114, opened 2026-04-07
Use localhost for Windows start script

Hurrieam:
Update package.json start:windows script to bind the dev server to 127.0.0.1 instead of 0.0.0.0. This restricts the Windows development server to localhost, preventing it from listening on all interfaces and reducing accidental external exposure.


chenglou:
Thanks! Binding to the LAN is intentional, so demos can be opened from phones and other devices, and `bun start` does the same. Closing, but thanks for the PR!

PR #124, opened 2026-04-11
Improve Korean keep-all mixed-script breaks

doyoon530:
## Summary

- Improves `wordBreak: keep-all` for Korean-heavy mixed-script tokens.
- Keeps Hangul + ASCII/digit product-like runs together, e.g. `AI정보공학과`, `README카드생성기`, `api문서v2가이드`, `2026학년도공지`.
- Stops keep-all continuation across path/query/key-value separators like `/`, `?`, `&`, `=`, and `:`. These separators are treated as structural boundaries so URL/path/query/key-value text does not get folded into one Korean token.
- Adds regression tests for compact Korean tokens, separator boundaries, punctuation, and ZWSP.

## Before / After

Before, ASCII-leading Korean tokens could stay split at the script boundary:

- `AI정보공학과` -> `AI` + `정보공학과`
- `2026학년도공지` -> `2026` + `학년도공지`

After, compact Korean mixed-script tokens stay together under keep-all, while structured separators remain break boundaries:

- `AI정보공학과` stays as one text unit
- `key=value한글` remains split around `=`
- `docs/README한글가이드` remains split around `/`

## Testing

- `bun run check`
- `bun test`

### DIFF
--- a/src/analysis.ts
+++ b/src/analysis.ts
@@ -98,9 +98,32 @@ export function setAnalysisLocale(locale?: string): void {
 const arabicScriptRe = /\p{Script=Arabic}/u
 const combiningMarkRe = /\p{M}/u
 const decimalDigitRe = /\p{Nd}/u
+// Korean app/product labels often mix Hangul with ASCII letters/digits and
+// lightweight token punctuation, but URL/query/key-value separators should
+// remain structural boundaries instead of being folded into one token.
+const keepAllTextRunSeparators = new Set(['/', '?', '&', '=', ':'])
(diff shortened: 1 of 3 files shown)


chenglou:
Thanks for this! I dug into the cases from the PR body and current main now matches our Chrome/Safari probes for the core Korean mixed-script keep-all cases, including the compact Hangul + ASCII/digit tokens. The remaining query-ish separator behavior looks browser-dependent enough that I don't want to take a broad change for it right now, so I'll close this rather than merge it. Appreciate the solid repros.

PR #125, opened 2026-04-12
docs: fix unclosed parenthesis in walkLineRanges description

hiSandog:
Fixed an unclosed parenthesis in the API glossary comment for `walkLineRanges`.\n\n**Before:**\n```\n(e.g. binary search ... is "nice" too. You can have text messages shrinkwrap and balanced text layout this way)\n```\n\n**After:**\n```\n(e.g. binary search ... is "nice" too). You can have text messages shrinkwrap and balanced text layout this way\n```\n\nSmall doc fix, no functional change.


chenglou:
Thanks! I folded this into cca3f1d and credited you as co-author.

PR #129, opened 2026-04-13
fix: preserve streamed line starts inside a chunk

### LINKED ISSUE #121: Bug: `layoutNextLine` and `walkLineRanges` mismatch with `layoutWithLines`
## Description

The chunked and non-chunked paths for streaming layout (`layoutNextLine`, `layoutNextLineRange`, and internal simple layout functions) contain multiple edge case bugs regarding how line starts are calculated when `whiteSpace` modes are altered (like `pre-wrap`) or when combining elements like `ZWSP`, spaces, and `soft-hyphens`. Additionally, these streaming layout functions drift out of sync with the batched `layoutWithLines` function when `simpleLineWalkFastPath` is disabled (e.g. `whiteSpace: 'pre-wrap'` or `wordBreak: 'keep-all'`).

## Reproducing the bug

Run the following code using `bun test_bug.ts`:

```ts
import { prepareWithSegments, layoutWithLines, layoutNextLine } from './src/layout.ts';

// Mock document and canvas for the reproduction to work outside a browser
globalThis.document = {
  body: { appendChild: () => {}, removeChild: () => {} },
  createElement: () => ({
    style: {},
    getContext: () => ({ measureText: (t) => ({ width: t.length * 10 }) }),
    getBoundingClientRect: () => ({width: 10}),
  })
} as any;

const text = "بام  \u200DB     bا \u00ADb\u060C b\f \u061F\uD83D\uDE80\u061F\u0639 \u0631 \u672C \u061F\na a A\u200B \u8A9E \u8A9E\u200D\u062D";
const font = "16px Inter";

const p1 = prepareWithSegments(text, font, {whiteSpace: 'normal', wordBreak: 'normal'}); // normal

const lwl1 = layoutWithLines(p1, 56.57, 20);

const streamingLines = [];
let cursor = {segmentIndex: 0, graphemeIndex: 0};
while (true) {
  const line = layoutNextLine(p1, cursor, 56.57);
  if (!line) break;
  streamingLines.push(line);
  cursor = line.end;
}

console.log("layoutWithLines vs layoutNextLine length mismatch:", lwl1.lines.length !== streamingLines.length);
```

## Root Cause

1. `normalizeSimpleLineStartSegmentIndex` and `normalizeLineStartInChunk` are hardcoded to skip past `'space'`, `'zero-width-break'`, and `'soft-hyphen'` by advancing `segmentIndex` unconditionally:
```typescript
  while (segmentIndex < prepared.widths.length) {
    const kind = prepared.kinds[segmentIndex]!
    if (kind !== 'space' && kind !== 'zero-width-break' && kind !== 'soft-hyphen') break
    segmentIndex++
  }
```
2. However, other whitespace configurations introduce different `kinds` (e.g. `'preserved-space'`, `'hard-break'`, `'tab'`) that these normalization functions fail to take into account properly. When `{ whiteSpace: 'pre-wrap' }` is on, spaces become `'preserved-space'` instead of `'space'`.
3. This skipping logic causes `layoutNextLine` to "eat" the wrong tokens when trying to find the beginning of the next line, or conversely, failing to skip things it *should* skip. This makes `layoutNextLine` start on the wrong segment/grapheme offset, causing infinite loops, dropping lines entirely, or missing/adding content.
4. Also, these functions do not accurately handle `zero-width-break` characters followed by spaces when determining where the next streamed line should begin, causing mismatch with the batch API.

lttlin:
## Problem
`layoutNextLine()` and `layoutNextLineRange()` renormalize every continuation cursor as if it were starting from a fresh paragraph or chunk boundary. In mixed-script edge cases with a `ZWSP` break opportunity followed by collapsible whitespace, that can skip a real continuation segment and merge two streamed lines that `layoutWithLines()` keeps separate.

## Root Cause
The line-start normalization helpers unconditionally skipped `space`, `zero-width-break`, and `soft-hyphen` segments even when the caller was resuming from an exact `end` cursor returned by the previous streamed line. That normalization is correct at the true start of a paragraph or chunk, but it is too aggressive for continuation cursors inside an existing chunk.

## Fix
Only trim collapsible leading break-opportunity segments when the cursor is at the true paragraph start or chunk start. If the caller resumes from a later in-chunk cursor, preserve it exactly and continue streaming from that segment.

## Validation
- `cd /tmp/pretext_issue_121 && ~/.bun/bin/bun test src/layout.test.ts --test-name-pattern 'streaming canary preserves exact continuation cursors after a ZWSP break opportunity|layoutWithLines strips leading collapsible space after a ZWSP break the same way as layoutNextLine|layoutNextLine reproduces layoutWithLines exactly'`
- `cd /tmp/pretext_issue_121 && ~/.bun/bin/bun test src/layout.test.ts`

Closes #121

### DIFF
--- a/src/layout.test.ts
+++ b/src/layout.test.ts
@@ -818,6 +818,30 @@ describe('layout invariants', () => {
     expect(actual).toEqual(expected.lines)
   })
 
+  test('streaming canary preserves exact continuation cursors after a ZWSP break opportunity', () => {
+    const text = 'بام  \u200DB     bا \u00ADb\u060C b\f \u061F🚀\u061Fع ر 本 \u061F\na a A\u200B 語 語\u200Dح'
+    const prepared = prepareWithSegments(text, FONT)
+    const width = 56.57
+    const expected = layoutWithLines(prepared, width, LINE_HEIGHT)
+
+    expect(collectStreamedLines(prepared, width)).toEqual(expected.lines)
+
+    const streamedRanges = []
+    let cursor = { segmentIndex: 0, graphemeIndex: 0 }
+    while (true) {
+      const line = layoutNextLineRange(prepared, cursor, width)
+      if (line === null) break
+      streamedRanges.push(line)
+      cursor = line.end
+    }
+
+    expect(streamedRanges).toEqual(expected.lines.map(line => ({
+      width: line.width,
+      start: line.start,
+      end: line.end,
+    })))
+  })
+
   test('layout and layoutWithLines stay aligned when ZWSP triggers narrow grapheme breaking', () => {
     const cases = [
       'alpha\u200Bbeta',
--- a/src/line-break.ts
+++ b/src/line-break.ts
@@ -42,6 +42,8 @@ function normalizeSimpleLineStartSegmentIndex(
   prepared: PreparedLineBreakData,
   segmentIndex: number,
 ): number {
+  if (segmentIndex > 0) return segmentIndex
+
   while (segmentIndex < prepared.widths.length) {
     const kind = prepared.kinds[segmentIndex]!
     if (kind !== 'space' && kind !== 'zero-width-break' && kind !== 'soft-hyphen') break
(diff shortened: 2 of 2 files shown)


chenglou:
Thanks! I've implement this and credited you as co-author

PR #133, opened 2026-04-14
Add Mouse & Cheese demo 

carobarbosaux:
A new demo page (/demos/mouse-cheese) showing Pretext's layoutNextLine() used for per-line variable-width layout: each story line gets a narrower available width while it overlaps the cheese block, so the text naturally wraps around the illustration. A slider resizes the cheese in real time and the text reflows without any DOM measurements in the resize path.

The cheese and mouse are inline SVGs (cute cartoon style). The demo is added to the demos index grid alongside the existing nine demos.

https://claude.ai/code/session_01P3CMU8eN1riSUg4EzCaMcA

### DIFF
--- a/pages/demos/index.html
+++ b/pages/demos/index.html
@@ -141,6 +141,11 @@
         <h2>Masonry</h2>
         <p>A text-card occlusion demo where height prediction comes from Pretext instead of DOM reads.</p>
       </a>
+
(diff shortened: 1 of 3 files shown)


chenglou:
Thanks for the fun demo! Text flowing around a moving shape is covered by the dynamic-layout and editorial-engine demos now, so I'll close this one. Appreciate it!

PR #138, opened 2026-04-16
simplify: unify stepRichInlineLine/Stats and remove containsCJKText wrapper

### LINKED ISSUE #136: Simplify rich-inline: unify stepRichInlineLine/Stats and remove containsCJKText wrapper
## Summary

Two small simplification opportunities in the v0.0.5 code, both pure refactors with no behavior change.

### 1. `stepRichInlineLine` and `stepRichInlineLineStats` share ~180 lines of near-identical logic

`rich-inline.ts:211–387` and `rich-inline.ts:389–519` implement the same line-stepping loop. The only difference is whether a `collectFragment` callback fires. Since `stepRichInlineLine` already accepts an optional `collectFragment` parameter, the stats variant could be replaced with `stepRichInlineLine(flow, maxWidth, cursor, undefined)`.

**Evidence of drift risk:** PR #132 had to apply the same overflow guard to both functions in parallel — the exact duplication pattern that makes one-sided patches easy to miss.

I verified the two functions produce identical `lineCount` and `maxLineWidth` across multiple item/width combinations (simple text, CJK multi-item, atomic chips with `extraWidth`, mixed items), so unification would not change behavior.

### 2. `containsCJKText()` is a single-call-site passthrough to `isCJK()`

```ts
// analysis.ts:161–163
function containsCJKText(text: string): boolean {
  return isCJK(text)
}
```

One call site at line 1214. Similar to the `isCJK` dedup and no-op removal in PR #118.

## Motivation

Both align with the repo's holistic-pass simplification principle. Neither changes behavior — the first reduces the surface for future one-sided patches in rich-inline, and the second removes a trivial indirection.

Happy to submit a PR if these look reasonable.

shaun0927:
Fixes #136

## Changes

Two small simplifications, bundled together since they are each one-liners with no cross-dependencies. Neither changes behavior — all 84 tests pass.

### 1. Remove `stepRichInlineLineStats` (~130 lines deleted)

`stepRichInlineLineStats` was a near-duplicate of `stepRichInlineLine` whose only difference was skipping the optional `collectFragment` callback. Since the callback is already optional (called via `collectFragment?.()`), passing `undefined` achieves the same result.

**Why this matters:** PR #132 had to apply the same overflow guard to both functions in parallel — the exact duplication pattern that makes one-sided patches easy to miss. Unifying the two functions ensures future fixes apply to both the materializing and stats-only paths automatically.

### 2. Inline `containsCJKText()` wrapper (4 lines deleted)

`containsCJKText()` was a single-call-site passthrough to the already-public `isCJK()`. Replaced the call at line 1210 with `isCJK(text)` directly. Similar to the `isCJK` dedup in PR #118.

## Test plan

- [x] `bun test src/layout.test.ts` — 84 tests pass, 0 fail
- [x] No public API changes — only internal function removal
- [x] `measureRichInlineStats` still returns identical results to `walkRichInlineLineRanges` across all existing test cases

### DIFF
--- a/src/analysis.ts
+++ b/src/analysis.ts
@@ -158,10 +158,6 @@ const keepAllGlueChars = new Set([
   '\uFEFF',
 ])
 
-function containsCJKText(text: string): boolean {
-  return isCJK(text)
-}
-
 function endsWithKeepAllGlueText(text: string): boolean {
   const last = getLastCodePoint(text)
   return last !== null && keepAllGlueChars.has(last)
@@ -1211,7 +1207,7 @@ function mergeKeepAllTextSegments(segmentation: MergedSegmentation): MergedSegme
     const start = segmentation.starts[i]!
 
     if (kind === 'text') {
-      const textContainsCJK = containsCJKText(text)
+      const textContainsCJK = isCJK(text)
       const textCanContinue = canContinueKeepAllTextRun(text)
 
       if (pendingTextParts !== null && pendingContainsCJK && pendingCanContinue) {
--- a/src/rich-inline.ts
+++ b/src/rich-inline.ts
@@ -386,139 +386,6 @@ function stepRichInlineLine(
   return lineWidth
 }
 
-function stepRichInlineLineStats(
-  flow: InternalPreparedRichInline,
-  maxWidth: number,
-  cursor: RichInlineCursor,
-): number | null {
-  if (flow.items.length === 0 || cursor.itemIndex >= flow.items.length) return null
-
-  const safeWidth = Math.max(1, maxWidth)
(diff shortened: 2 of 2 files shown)


chenglou:
Thanks for the PR. Landed the useful cleanup on main in ad2bc4e and credited you as co-author. I kept the core idea: one rich-inline line stepper instead of a stats-only clone, plus the tiny `containsCJKText()` wrapper removal. I adjusted the implementation so `measureRichInlineStats()` keeps the markdown-chat hot-path shape, while `layoutNextRichInlineLineRange()` still clones the public start cursor; added a regression test for that too. Closing this as covered.

PR #139, opened 2026-04-16
docs: document breakableFitAdvances cache invariant

### LINKED ISSUE #137: getSegmentBreakableFitAdvances: cached result does not distinguish BreakableFitMode
## Summary

`getSegmentBreakableFitAdvances()` in `measurement.ts:196–264` accepts a `mode: BreakableFitMode` parameter (`sum-graphemes`, `segment-prefixes`, or `pair-context`), but the cache check at line 203 returns the stored result without verifying the mode matches:

```ts
if (metrics.breakableFitAdvances !== undefined) return metrics.breakableFitAdvances
// mode is not consulted
```

If the same segment text were ever requested with two different modes, the first-cached result would silently serve both.

## Why it is safe today

The mode is fully determined by text content and the session-invariant engine profile:

1. `isNumericRunSegment(text)` is true → always `pair-context`
2. `engineProfile.preferPrefixWidthsForBreakableRuns` is true → always `segment-prefixes`
3. Otherwise → always `sum-graphemes`

Because `isNumericRunSegment` is a pure text predicate and the engine profile does not change within a session, the same segment text always arrives with the same mode. I verified this by calling `getSegmentBreakableFitAdvances` with different modes on the same segment — the cache returns the identical object reference, confirming the mode is ignored, but the current call pattern prevents a mismatch from occurring.

## Suggestion

The function's type signature promises mode-awareness that the implementation does not deliver. One of:

1. **Assert:** add `assert(cachedMode === mode)` guarded by `process.env.NODE_ENV !== 'production'` — zero production cost, catches future regressions
2. **Document:** add a comment at the cache check noting the invariant and why it holds
3. **Refactor:** remove `mode` from the signature entirely and let the caller compute advances externally, making the cache contract explicit

Not filing this as a bug — the invariant holds — but the gap between the API surface and the implementation could surprise a future caller.

shaun0927:
Fixes #137

## Problem

`getSegmentBreakableFitAdvances()` accepts a `mode: BreakableFitMode` parameter but the cache at line 203 returns the stored result without checking whether the mode matches. If a future caller requested the same segment text with a different mode, the cache would silently return the wrong advances.

## Why the cache is safe today

The mode is fully determined by two things that are constant for a given segment text within a session:

1. `isNumericRunSegment(text)` — pure text predicate, always returns the same result for the same string → forces `pair-context`
2. `engineProfile.preferPrefixWidthsForBreakableRuns` — detected once from `navigator.userAgent`, invariant for the session → forces `segment-prefixes`
3. Otherwise → `sum-graphemes`

So the same segment text always arrives with the same mode, and the mode-agnostic cache never returns a mismatched result.

## Change

Added a 5-line comment above the cache check explaining the invariant and the consequence of breaking it. No behavior change, no public API change. All 84 tests pass.

## Test plan

- [x] `bun test src/layout.test.ts` — 84 pass, 0 fail
- [x] Comment-only change — no runtime effect

### DIFF
--- a/src/measurement.ts
+++ b/src/measurement.ts
@@ -200,6 +200,11 @@ export function getSegmentBreakableFitAdvances(
   emojiCorrection: number,
   mode: BreakableFitMode,
 ): number[] | null {
+  // The cached result is mode-agnostic: the current call sites guarantee that
+  // a given segment text always arrives with the same mode, because mode is
+  // determined by isNumericRunSegment(text) (text-deterministic) and the
+  // session-invariant engine profile.  If a future caller breaks this
+  // invariant, the cache will silently return the wrong advances.
   if (metrics.breakableFitAdvances !== undefined) return metrics.breakableFitAdvances
 
   const graphemeSegmenter = getSharedGraphemeSegmenter()


chenglou:
Thanks for the careful PR. I left the fuller context over here: https://github.com/chenglou/pretext/issues/137#issuecomment-4403238549 — main already handles this by making the cache mode-aware and pinning it with a regression test, so closing this without merging.

PR #140, opened 2026-04-17
Perf/streaming layout optimizations

sid-sadhana:


chenglou:
Thanks for this. I didn’t end up taking the side-table approach here, but this did nudge me to revisit the streaming chunk lookup path. I landed a smaller internal cleanup in 3757fa0 that keeps the public API unchanged, removes the duplicate chunk lookup in layoutNextLine() / layoutNextLineRange(), and improves the chunk-heavy benchmark. Appreciate the pointer.

Issue #143, opened 2026-04-17
Add direct tests for documented public APIs


nice-hang:
Some documented public APIs are exported but do not have clear direct regression
   coverage in src/layout.test.ts.

  Notably:
  - materializeLineRange()
  - measureNaturalWidth()
  - layoutNextRichInlineLineRange()

  We already cover related behavior indirectly, but adding small direct tests for
  these entry points would make the public API surface safer to refactor and keep
  docs/exports/tests aligned.

  Suggested scope:
  - add focused tests in src/layout.test.ts
  - reuse existing test patterns and fake canvas setup
  - keep this test-only with no production code changes

chenglou:
Thanks for the useful test-surface callout. Covered on main in 0355ea9 with focused direct tests for `materializeLineRange()` and `measureNaturalWidth()`.

PR #149, opened 2026-04-18
test: add direct coverage for documented public APIs

nice-hang:
## Summary

  - add direct regression tests for materializeLineRange()
  - add direct regression tests for measureNaturalWidth()
  - add direct regression tests for layoutNextRichInlineLineRange()

## Test plan
`bun test src/layout.test.ts`
image

## Related issue
https://github.com/chenglou/pretext/issues/143


chenglou:
Thanks for the PR! I landed the useful coverage on main in 0355ea9 and credited you as co-author. I trimmed the patch so we keep the direct public-helper coverage without duplicating an existing rich-inline wrapper/materialization check. Closing this as covered.

PR #156, opened 2026-04-21
feat: letterSpacing option via Canvas native API

mayrang:
## Summary

Adds `letterSpacing?: string` to `PrepareOptions`. Instead of computing letter-spacing mathematically (as in #108), this sets `ctx.letterSpacing` on the Canvas context before `measureText()`, letting the browser engine handle exact width calculations natively.

```ts
const prepared = prepare(text, font, { letterSpacing: '2px' })
const { height } = layout(prepared, 300, 20)
```

I'd appreciate feedback on whether this direction aligns with what you had in mind when you noted "Leaving this open for actual support work" in cca3f1d.

## Approach

**Why Canvas native (`ctx.letterSpacing`, Baseline 2025.03):**
- `measureText()` automatically includes letter-spacing for all segment types — CJK, emoji, ligatures, relative units like `em`
- No grapheme counting, no trailing-trim edge cases
- Measurement concern stays in `measurement.ts`; `layout()` hot path and `line-break.ts` are untouched

**Why `letter-spacing` only:**
- `font-optical-sizing`, `font-feature-settings`, and `font-variation-settings` are not in the Canvas 2D spec
- The only alternative for those would be DOM fallback measurement, which contradicts the library's design
- Partially addressing #107 seemed better than blocking on the full set

## What changed

| File | Change |
|------|--------|
| `src/measurement.ts` | `ctx.letterSpacing` before measurement, cache key `:ls=<value>`, emoji correction with letterSpacing |
| `src/layout.ts` | `PrepareOptions` type + parameter threading (5 lines) |
| `src/layout.test.ts` | Mock Canvas update + 6 unit tests |
| `README.md` | Documentation update |

## Accuracy

I ran the browser accuracy sweep locally with letter-spacing variants (`1px`, `0.05em`, `-0.5px`) added:

|  | Chrome | Safari |
|--|--------|--------|
| **Base sweep** | 7680/7680 (100%) | 7680/7680 (100%) |
| **Letter-spacing sweep** | 22955/23040 (99.63%) | 22992/23040 (99.79%) |

The base sweep is completely unaffected. The letter-spacing mismatches are:

- **Arabic/RTL text** (84 in Chrome, 48 in Safari): The existing Canvas-vs-DOM Arabic shaping gap (currently absorbed by `lineFitEpsilon`) is amplified when letter-spacing widens per-segment widths past the epsilon boundary. Not new measurement bugs — the same fine-width field, made more visible by letter-spacing.

- **1 Latin overflow-wrap** (Verdana 24px, ls=1px, w=500, Chrome only): Cumulative rounding in the existing `sum-graphemes` `breakableFitAdvances` mode. Canvas prefix measurement and DOM agree on the break point; only the grapheme-sum path diverges.

I did not modify the accuracy sweep infrastructure — happy to add letter-spacing coverage to the sweep if you'd like, in whatever structure makes sense.

## Open questions

1. **Is Canvas native the right approach?** The alternative (PR #108's mathematical approximation) avoids the `ctx.letterSpacing` API dependency but has correctness issues at line boundaries. Happy to adjust if you prefer a different direction.

2. **Should `breakableFitAdvances` mode selection change when letterSpacing is active?** The one Latin mismatch would be resolved by using `segment-prefixes` instead of `sum-graphemes` for letter-spaced text, but that's a policy change I didn't want to make without input.

## Test plan

- [x] 94 unit tests pass (88 existing + 6 new letterSpacing tests)
- [x] TypeScript compiles cleanly (`ctx.letterSpacing` available in lib target)
- [x] Base accuracy sweep: 7680/7680 (100%) in Chrome and Safari — zero regression
- [x] `layout()` hot path, `line-break.ts`, `analysis.ts` unchanged

Partially addresses #107 (letter-spacing only).

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,6 +241,8 @@ Pretext doesn't try to be a full font rendering engine (yet?). It currently targ
   const prepared = prepare(text, font, { letterSpacing: '2px' })
   ```
 
+  The value is a CSS `<length>` string (e.g. `'1px'`, `'0.05em'`). Invalid values are silently ignored by the Canvas API, falling back to no spacing.
+
   Other CSS text features outside the canvas `font` shorthand, such as `font-optical-sizing`, `font-feature-settings`, and standalone `font-variation-settings`, are not modeled separately. These properties are not in the Canvas 2D spec. Variable-font axes only help when the active axis is reflected in the canvas font string, for example via weight.
 
 ## Develop
--- a/README.md
+++ b/README.md
@@ -235,7 +235,13 @@ Pretext doesn't try to be a full font rendering engine (yet?). It currently targ
 - `{ wordBreak: 'keep-all' }` is supported too. It behaves like you'd expect for CJK/Hangul text, while keeping the same `overflow-wrap: break-word` fallback for overlong runs.
 - `system-ui` is unsafe for `layout()` accuracy on macOS. Use a named font.
 - Runtime requires `Intl.Segmenter` and Canvas 2D text measurement. Browsers or runtimes without `Intl.Segmenter` are currently unsupported.
-- CSS text features outside the canvas `font` shorthand, such as `letter-spacing`, `font-optical-sizing`, `font-feature-settings`, and standalone `font-variation-settings`, are not modeled separately. Variable-font axes only help when the active axis is reflected in the canvas font string, for example via weight.
+- `letter-spacing` is supported via the `letterSpacing` option in `prepare()`:
+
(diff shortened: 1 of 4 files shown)


chenglou:
Thanks for exploring the native Canvas `ctx.letterSpacing` approach and the accuracy tradeoffs. Numeric px `letterSpacing` support landed on `main` in 1a8b2ae. Credited your PR!

PR #160, opened 2026-04-28
chore: mark package as sideEffects: false to enable tree-shaking

### LINKED ISSUE #158: Add "sideEffects": false to enable tree-shaking
### Context

I maintain [`@darkroomengineering/fitbox`](https://github.com/darkroomengineering/fitbox), a text-fitting library built on pretext. We import only four symbols from the main entry:

```ts
import {
  prepareWithSegments,
  measureNaturalWidth,
  measureLineStats,
  layoutWithLines,
} from '@chenglou/pretext';
```

### Observation

`package.json` doesn't declare `"sideEffects": false`, so downstream bundlers (Vite/webpack/rollup) conservatively ship the full transitive closure of pretext's internal modules:

| File | Size (raw) |
|---|---:|
| `analysis.js` | 37 KB |
| `line-break.js` | 33 KB |
| `generated/bidi-data.js` | 23 KB |
| `layout.js` | 22 KB |
| `measurement.js` | 8 KB |
| `bidi.js` | 6 KB |
| **Total** | **~129 KB raw** |

Every downstream consumer ships all 129 KB, even when they exercise only a fraction of the exported surface.

### Ask

```json
{ "sideEffects": false }
```

A scan of `dist/*.js` shows only top-level `const`/`let` initialization (regexes, `Set`s, `let sharedWordSegmenter = null`). All pure expressions, safe to mark side-effect-free. With the flag, bundlers can drop unused exports based on what consumers actually import (e.g., the seven `layout.d.ts` exports fitbox doesn't use, plus unused helpers in `analysis.js`/`line-break.js`).

Realistic impact: 20–40% bundle reduction for consumers using a subset of the API, which is most of them.

### Secondary (lower priority, separate issue if interested)

`generated/bidi-data.js` is 23 KB of Unicode data needed only for RTL text. If `bidi.js` could lazy-load it (dynamic `import()` gated on a quick RTL scan), latin-only sites would drop 23 KB automatically. Not worth tackling until the `sideEffects` change is measured.

### Willing to PR

Happy to open a one-line PR for the `sideEffects` change, plus a small bundle-size regression test if that's welcome.

mvanhorn:
## Summary

Add `"sideEffects": false` to package.json so downstream bundlers can tree-shake unused exports of `@chenglou/pretext`.

## Why this matters

Per [#158](https://github.com/chenglou/pretext/issues/158), the `darkroomengineering/fitbox` maintainer's bundle audit shows ~129 KB of pretext modules ship to every consumer regardless of which exports they actually import:

| File | Size (raw) |
|---|---:|
| `analysis.js` | 37 KB |
| `line-break.js` | 33 KB |
| `generated/bidi-data.js` | 23 KB |
| `layout.js` | 22 KB |
| `measurement.js` | 8 KB |
| `bidi.js` | 6 KB |
| **Total** | **~129 KB raw** |

A consumer importing only `prepareWithSegments`, `measureNaturalWidth`, `measureLineStats`, `layoutWithLines` ships the full set today.

A scan of the published modules confirms there are no top-level side effects: `let sharedWordSegmenter = null` and the regex/Set initializers are pure expressions, evaluated lazily on first call. Marking the package side-effect-free is safe and lets bundlers drop unused exports based on what consumers import.

Realistic impact per #158: 20-40% bundle reduction for consumers using a subset of the API.

## Changes

One line in `package.json`:

```json
"sideEffects": false
```

## Testing

- `bun run build:package` — clean build
- `bun run check` — `tsc && oxlint --type-aware src` reports 0 warnings, 0 errors
- `bun test` — 103 pass, 0 fail (417 expect calls)

No source code touched, no behavior change.

Fixes #158

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "url": "git+https://github.com/chenglou/pretext.git"
   },
   "type": "module",
+  "sideEffects": false,
   "publishConfig": {
     "access": "public"
   },


chenglou:
Thanks for the PR! Landed this on main in dea8592, with the small export cleanup so demos/assets remain installed as examples rather than direct package imports. Closing this as covered.

PR #161, opened 2026-04-30
fix: cache breakable fit advances per mode

### LINKED ISSUE #137: getSegmentBreakableFitAdvances: cached result does not distinguish BreakableFitMode

hiSandog:
## Summary
- cache breakable fit advances separately for each fit mode
- preserve the long segment-prefix fallback under the pair-context cache slot
- add a regression test covering repeated calls across sum, pair, and prefix modes

Fixes #137.

## Validation
- npx tsc --noEmit
- npx oxlint --type-aware src
- git diff --check

Note: bun is not installed in this environment, so I could not run bun test src/layout.test.ts.

### DIFF
--- a/src/layout.test.ts
+++ b/src/layout.test.ts
@@ -12,6 +12,8 @@ type LayoutModule = typeof import('./layout.ts')
 type LineBreakModule = typeof import('./line-break.ts')
 type RichInlineModule = typeof import('./rich-inline.ts')
 type AnalysisModule = typeof import('./analysis.ts')
+type MeasurementModule = typeof import('./measurement.ts')
+type SegmentMetrics = ReturnType<MeasurementModule['getSegmentMetrics']>
 
 let prepare: LayoutModule['prepare']
 let prepareWithSegments: LayoutModule['prepareWithSegments']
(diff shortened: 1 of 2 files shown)


chenglou:
Thanks for the PR! Main already has the correctness fix from 1a8b2ae, and I added the targeted regression coverage in 2adcbc8. I’m going to close this without merging since the multi-mode cache implementation isn’t needed right now.

PR #164, opened 2026-05-03
Add guard for Intl.Segmenter in line-text.ts, Fixes #159

### LINKED ISSUE #159: Intl.Segmenter hard dependency crashes on Firefox Mobile <125 without documentation or guard
Hi! We’re using `@chenglou/pretext@0.0.5` for chord-sheet layout and ran into a `TypeError: Intl.Segmenter is not a constructor thrown from getSharedGraphemeSegmenter` on Firefox Mobile 110 / Android. Firefox shipped `Intl.Segmenter` in 125 (Apr 2024), so pre-125 builds (which still exist in the long tail — custom GeckoView embeds, locked Play Store devices) crash as soon as `prepare()` is reached.

Treating `Intl.Segmenter` as a baseline is totally reasonable given how much it saves for CJK/Thai/Arabic. The issues are that (a) there’s no upfront guard, so the failure surfaces as an obscure minified `TypeError` deep in a render stack, and (b) the requirement isn’t in the `README` or `engines`. We shipped `@formatjs/intl-segmenter` as a conditional polyfill on our side, but wanted to flag this for others.

Would you consider:

A module-load typeof `Intl.Segmenter` check that throws with a clear message + polyfill pointer?
A note in the README (or `browserslist`) documenting the requirement?
Optional: an injectable segmenter factory so integrators can plug in a polyfill per instance instead of patching globals?

ArjunSrivastava1:
Fixes #159 

- Adds a runtime guard that throws an error message when Intl.Segmenter is unavailable, preventing from finding the issue  deep in a render stack somewhere.
- The error also consists of the [site](https://caniuse.com/mdn-javascript_builtins_intl_segmenter) which lets the developers check if they are able to run the segmenter or not.
- Further, this fix includes a comment that also mentions the reqs so code and explains so its clearer.

It was deemed unecessary to add the same in the readme as it felt bloated so that wasnt done

> [!WARNING]
> **GitHub PR merge issues (since April 23, 2026).**
> A major issue is currently ongoing in github where the merge button is not working properly, confirmed since 23 April 2026, the issue is that merging leads to rolling back of commits, loss of history, and wrong merges which leads to loss of data 
> Be careful merging anything and always double check

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "markdown.validate.enabled": true
+}
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,0 +1,342 @@
+lockfileVersion: '9.0'
+
+settings:
+  autoInstallPeers: true
+  excludeLinksFromLockfile: false
+
+importers:
+
+  .:
+    devDependencies:
+      '@types/bun':
+        specifier: latest
+        version: 1.3.13
+      marked:
+        specifier: ^17.0.5
+        version: 17.0.6
+      oxlint:
+        specifier: ^1.51.0
+        version: 1.62.0(oxlint-tsgolint@0.15.0)
+      oxlint-tsgolint:
+        specifier: ^0.15.0
+        version: 0.15.0
+      tsgolint:
+        specifier: ^0.0.1
+        version: 0.0.1
+      typescript:
+        specifier: 6.0.2
+        version: 6.0.2
+
+packages:
+
+  '@oxlint-tsgolint/darwin-arm64@0.15.0':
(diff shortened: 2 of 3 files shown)


chenglou:
Thanks for taking a pass at this. This doesn’t polyfill `Intl.Segmenter`; it adds a clearer lazy error in one helper, plus a couple unrelated local files. Since the requirement is now documented in the README and we’re not adding runtime polyfill plumbing or partial guards right now, I’m going to close this without merging.

ArjunSrivastava1:
yeah, that makes sense - i liked taking a stab at it, adding a note in the readme is good enough for this as of now instead of adding plumbing, i just wanted to see if a clearer error could work for this as well
ur very welcome

PR #165, opened 2026-05-04
fix: keep German low quotes with following word

### LINKED ISSUE #163: German low opening quotes (‚ U+201A, „ U+201E) missing from kinsokuEnd
First off — thank you for pretext. We've been using it for the layout math behind a tracked-changes review surface, and getting accurate line breaks and per-fragment X offsets out of one library has been genuinely lovely. Really appreciate the care that's gone into this.

## The issue

The `kinsokuEnd` set in `dist/analysis.js` (line-end-prohibited / forward-
sticky opening punctuation) contains the standard openers:

```js
const kinsokuEnd = new Set([
    '"',
    '(', '[', '{',
    '“', '‘', '«', '‹',
    // ...CJK openers
]);
```

…but it's missing the German low opening quotes:

- `‚` U+201A SINGLE LOW-9 QUOTATION MARK — German opening single quote
- `„` U+201E DOUBLE LOW-9 QUOTATION MARK — German opening double quote
In German typography, „Wort" and ‚Wort' open with the low forms, and they should not end a line — exactly the role `kinsokuEnd` is meant to capture. Browsers (Chromium in our testing) keep them glued to the following word; pretext breaks between them.

## Repro

```js
import {
  prepareRichInline,
  layoutNextRichInlineLineRange,
  materializeRichInlineLineRange,
} from "@chenglou/pretext/rich-inline";

function lines(text, width) {
  const prepared = prepareRichInline([{ text, font: "16px sans-serif" }]);
  const out = [];
  let cursor;
  for (;;) {
    const range = layoutNextRichInlineLineRange(prepared, width, cursor);
    if (!range) break;
    out.push(materializeRichInlineLineRange(prepared, range)
              .fragments.map(f => f.text).join(""));
    cursor = range.end;
  }
  return out;
}

const W = 180;
console.log("„", lines("aaaaaaaaaaaaaaaaaaa „Wort", W));
console.log("‚", lines("aaaaaaaaaaaaaaaaaaa ‚Wort", W));
console.log("“", lines("aaaaaaaaaaaaaaaaaaa “Wort", W));
console.log("‘", lines("aaaaaaaaaaaaaaaaaaa ‘Wort", W));
```

Observed output:

```bash
„ [ 'aaaaaaaaaaaaaaaaaaa „', 'Wort' ]   ← quote stranded on line 1
‚ [ 'aaaaaaaaaaaaaaaaaaa ‚', 'Wort' ]   ← quote stranded on line 1
“ [ 'aaaaaaaaaaaaaaaaaaa ',  '“Wort' ]  ← stays glued (correct)
‘ [ 'aaaaaaaaaaaaaaaaaaa ',  '‘Wort' ]  ← stays glued (correct)
```

Chromium, given the same string and width: 180px (verified with `Range.getClientRects()`), starts line 2 with the quote in all four cases — i.e. the English curly quotes match Chromium, and the German low quotes don't.

## Suggested fix

Adding the two codepoints to `kinsokuEnd` should be enough:

```js
const kinsokuEnd = new Set([
    '"',
    '(', '[', '{',
    '“', '‘', '«', '‹',
    '‚', '„',         // U+201A, U+201E — German opening quotes
    // ...CJK openers
]);
```

The matching closers `’` U+2019 and `”` U+201D are already in `leftStickyPunctuation`, so the asymmetry is what made this easy to miss — German just happens to spell the openers with different codepoints than the closers.

### Note on UAX #14

Both characters carry line-break class QU (Ambiguous Quotation), which UAX #14 §7 explicitly calls out as locale-dependent — they behave as openers in German and as closers in some other locales. A locale-agnostic addition to kinsokuEnd matches what Chromium does in practice across the locales we've tested, but if you'd rather gate this behind a German-specific tailoring (similar to how some libraries expose a language: "de" option), that would also work for us — we'd just call into whichever API you settle on.

Thanks again for the amazing library!

syahmiharith:
## Summary
- Add German low opening quotes `‚` and `„` to `kinsokuEnd` so they stay attached to the following word.
- Add a regression test that verifies the low quote is not stranded at the end of a line.

## Validation
- `bun test`
- `bun run check`
- `bun run build:package`

Fixes #163

### DIFF
--- a/src/analysis.ts
+++ b/src/analysis.ts
@@ -221,7 +221,7 @@ export const kinsokuStart = new Set([
 export const kinsokuEnd = new Set([
   '"',
   '(', '[', '{',
-  '“', '‘', '«', '‹',
+  '“', '‘', '«', '‹', '‚', '„',
   '\uFF08',
   '\u3014',
   '\u3008',
--- a/src/layout.test.ts
+++ b/src/layout.test.ts
@@ -445,6 +445,19 @@ describe('prepare invariants', () => {
     expect(prepared.segments).toEqual(['“Whenever'])
   })
 
+  test('keeps German low opening quotes attached to the following word', () => {
+    for (const quote of ['‚', '„']) {
+      const prepared = prepareWithSegments(`aaaaaaaaaaaaaaaaaaa ${quote}Wort`, FONT)
+      expect(prepared.segments).toEqual(['aaaaaaaaaaaaaaaaaaa', ' ', `${quote}Wort`])
+
+      const strandedQuoteWidth = measureWidth(`aaaaaaaaaaaaaaaaaaa ${quote}`, FONT) + 0.1
+      expect(layoutWithLines(prepared, strandedQuoteWidth, LINE_HEIGHT).lines.map(line => line.text)).toEqual([
+        'aaaaaaaaaaaaaaaaaaa ',
+        `${quote}Wort`,
+      ])
+    }
+  })
+
   test('keeps apostrophe-led elisions attached to the following word', () => {
     const prepared = prepareWithSegments('“Take ’em downstairs', FONT)
     expect(prepared.segments).toEqual(['“Take', ' ', '’em', ' ', 'downstairs'])


chenglou:
Thanks for the PR! Landed this on main in 696a414 with the same core fix, plus the other browser-confirmed inverted openers (`¡`, `¿`, `⸘`). Closing this as covered by the mainline commit.

PR #166, opened 2026-05-04
chore: support Chrome browser checks on Windows

syahmiharith:
## Summary
- Add a non-macOS Chrome automation path that launches headless Chrome through the DevTools Protocol instead of AppleScript.
- Start the internal Bun page server directly through the current Bun executable instead of `/bin/zsh`.
- Use the platform temp directory for browser automation locks and document the `CHROME_BIN` / `CHROME_PATH` override.

## Why
Trying to follow the current TODO canary workflow on Windows failed before any browser evidence could be collected:

- lock creation tried to use a Unix-style temp fallback
- Chrome automation called `osascript`
- page serving depended on `/bin/zsh`

This keeps the existing macOS Chrome/Safari behavior intact while making Chrome-based checks runnable on Windows.

## Validation
- `bun run check`
- `bun run corpus-check --id=mixed-app-text 710 --diagnose`
- `bun run corpus-check --id=mixed-app-text 710 --diagnose --method=span`
- `bun run corpus-taxonomy --id=mixed-app-text 710`
- `bun run probe-check --text="hello world from pretext" --width=160`

## Notes
The mixed app canary at 710px is height/line-count exact in this Windows Chrome run. The diagnostic still reports a one-character break-opportunity difference around `価格は¥12,800です`, so I did not make an engine change from this evidence.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -41,6 +41,7 @@ bun install
   On a first-break mismatch, probe output now includes a small break trace.
   `sN:gM` means segment/grapheme position, `unit` is that unit's width, `fit` is the cumulative fitted width from the current line start, and `[ours]` / `[browser]` mark the competing break boundaries.
   For Safari URL/query misses or other extractor-sensitive cases, cross-check `--method=span` before changing the engine.
+  On non-macOS hosts, Chrome checks launch headless Chrome through DevTools; set `CHROME_BIN` or `CHROME_PATH` if Chrome is not in a standard location.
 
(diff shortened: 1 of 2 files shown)


chenglou:
Thanks @syahmiharith! I folded your platform temp-directory fix into [1c53c4c](https://github.com/chenglou/pretext/commit/1c53c4c5348f705b0abf1a51735ee3ecca0af9dd), with co-author credit. Appreciate you tracking down the Windows setup problems.

Issue #173, opened 2026-05-08
Support rich inline with `pre-wrap` whitespace preserve


mordechaim:
Rich inline allows mixed formatted content, but it explicitly says that pre-wrap is not supported.

Is there a reason we can't support whitespace preservation? I'm working on a Canva style editor and I have a need for rich inline with whitespace.

Stubenhocker1399:
Just noting that there's a related pr open in https://github.com/chenglou/pretext/pull/193

chenglou:
Makes sense for an editor, and it's not a hard no. Preserved spaces and tabs across styled pieces hang and wrap in ways I'd need to check against real browsers before promising anything. Keeping this open while I decide whether to support it.

ChuanfengZhang:
Thanks for clarifying. We have run into the same issue in a production editor: the paragraph is rendered with `white-space: pre-wrap`, but marks and font changes split it into multiple styled pieces. If the rich-inline path collapses consecutive spaces at those boundaries, the measurement can end up one line short and throw off pagination.

We currently carry a downstream patch that adds `prepareRichInline(items, { whiteSpace: "pre-wrap" })`. We will not reopen an implementation PR based on that patch. If useful, we can provide a minimal reproduction from the editor and native line-break observations from the browsers in whatever format works best for you.

PR #193, opened 2026-05-29
feat: support whiteSpace: 'pre-wrap' in prepareRichInline

ChuanfengZhang:
This allows rich text items to preserve leading and trailing spaces, mirroring the behavior of white-space: pre-wrap in CSS. Includes tests for the new mode and ensures backward compatibility with the default 'normal' mode.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ This usage allows rendering to canvas, SVG, WebGL and (eventually) server-side. 
 
 For hyphenation in manual layout, insert soft hyphens before `prepare()` / `prepareWithSegments()`. Pretext treats them as optional break points: unchosen soft hyphens stay invisible, while chosen breaks materialize as a trailing `-`. For mixed-language or user-generated app text, prefer conservative, locale-aware insertion over aggressive pattern hyphenation. Automatic hyphenation is not built in today.
 
-If your manual layout needs a small helper for rich-text inline flow, code spans, mentions, chips, and browser-like boundary whitespace collapse, there is a helper at `@chenglou/pretext/rich-inline`. It stays inline-only and `white-space: normal`-only on purpose:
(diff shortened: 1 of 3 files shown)


Stubenhocker1399:
Is it expected that with these changes calls to `measureRichInlineStats` with a maxWidth set can be overflowed by long running sequences of spaces?

ChuanfengZhang:
> Is it expected that with these changes calls to `measureRichInlineStats` with a maxWidth set can be overflowed by long running sequences of spaces?

Yes.

Stubenhocker1399:
Would be really useful to support this without overflowing the set maximum width, as otherwise setting a maximum width loses its purpose.

12joan:
I'm not sure trailing white space on inline items is handled correctly at the moment even for the default `white-space: normal` mode.

In HTML, `<u>Hello </u>world` is rendered with an underline under the space, as you would expect. Pretext strips this space and replaces it with a `gapBefore` on `world`, so there's no underline on the space.

In addition, when rendering this Pretext output using HTML, the lack of a space character has a negative impact on accessibility and usability. For example, if you copy the content of https://chenglou.me/pretext/rich-note/ as plain text, you get this:

> Ship@maya'srich-notecard oncepre-wraplands. Status
> blockedbyvertical textresearch, but 北京 copy and Arabic QA 
> are both green ✅. KeepجاهزforCmd+Kdocs; the review bundle 
> now includes 中文 labels, عربي fallback, and one more launch 
> pass 🚀 forFri 2:30 PM. KeeplayoutNextLine()public, tag this
> P1, keep3 reviewers, and route feedback todesign sync.

I'm not sure what the solution is, but it would probably be worth taking this into consideration before making any changes to how white space is handled.

chenglou:
Thanks for this, and sorry for the slow reply! The rich-inline helper has been reworked twice since, so this can't rebase cleanly. Pre-wrap across items also needs a proper design first. Closing this, but #173 stays open for the discussion. Thanks!

Issue #177, opened 2026-05-16
`rich-inline` wraps punctuation inconsistently with browser


rf-:
In the browser, if adjacent characters would stay attached during wrapping within the same node, they'll also stay attached when they're split between adjacent inline nodes. It looks like pretext might treat all item boundaries as breakable:

<img width="140" height="237" alt="Image" src="https://github.com/user-attachments/assets/c889464d-568c-48c6-8899-4edd08abebfe" />

Repro CodePen here: https://codepen.io/editor/rf3521dcb4b0/pen/019e31dd-8759-7933-a667-aa946717c2d2

hiSandog:
This seems like a good regression test to phrase as "same text, same wrap opportunities" across a single inline node versus adjacent inline nodes. In other words, `foo,` in one node and `foo` + `,` split across rich-inline boundaries should produce the same break behavior.

That framing would also help separate Unicode line-breaking behavior from rendering details: item boundaries should not become extra soft break opportunities unless the browser would allow one at the same character boundary.

chenglou:
Fixed in #241 for Chrome and Safari: rich-inline items now break only where their joined text breaks, so punctuation at the start of an item stays with the word before it. Firefox still breaks at every item boundary for now.

PR #194, opened 2026-05-31
Fix rich-inline punctuation boundary wrapping

luyh7:
## Summary
- Keep rich-inline item boundaries from wrapping before line-start-prohibited punctuation when there is no collapsed whitespace gap.
- Reuse the core punctuation sets so split inline items such as `Hello` + `.` stay attached.
- Add a regression test covering materialized rich-inline lines and stats.

## Why
Adjacent inline items should not introduce an extra line break before punctuation that browsers keep attached to the previous text node. This addresses #177.

## Validation
- `/home/luyh7/tmp/.tools/bun-1.3.14/package/bin/bun test src/layout.test.ts`
- `node node_modules/typescript/bin/tsc --noEmit`
- `node node_modules/typescript/bin/tsc -p tsconfig.build.json`
- `git diff --check`

### DIFF
--- a/src/layout.test.ts
+++ b/src/layout.test.ts
@@ -960,6 +960,30 @@ describe('rich-inline invariants', () => {
     })
   })
 
+  test('rich inline keeps line-start punctuation attached across item boundaries', () => {
+    const prepared = prepareRichInline([
+      { text: 'Hello', font: FONT },
+      { text: '.', font: FONT },
+      { text: ' world', font: FONT },
+    ])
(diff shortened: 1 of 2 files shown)


rf-:
If it's true that browsers behave exactly the same way for adjacent spans that they do for a single text node, it seems like doing this correctly would require a much bigger refactor so that all of the logic in `buildMergedSegmentation` can be applied across items. That's pretty difficult though—I'm taking a stab at it but haven't fully proven it out yet, and it's hard to not regress on performance at all in the simple case.

PR #197, opened 2026-06-04
docs: add Claude Code usage skill for pretext

GuyNachshon:
## What

Adds `.claude/skills/pretext/SKILL.md` — a [Claude Code](https://claude.com/claude-code) *skill* that teaches the assistant how to **use** `@chenglou/pretext` correctly. Since the repo already ships a `CLAUDE.md`, this lets anyone working in the repo with Claude Code get accurate, on-demand guidance without re-reading the full README each time.

It's a single Markdown file, no code or build changes.

## Why a skill instead of more README

The README is (rightly) the exhaustive API reference. A skill is a thin **decision/usage layer** on top of it — it loads only when relevant and focuses on the judgment calls the reference doesn't foreground:

- **Leads with the `prepare`-once / `layout`-many model** and the "never re-`prepare()` the same `(text, font, options)`" rule, since most correct usage follows from respecting that seam.
- **A decision guide** for picking between the three API families: fast path (`prepare`/`layout`), rich manual layout (`prepareWithSegments` + `walkLineRanges`/`layoutNextLineRange`/`layoutWithLines`), and the `rich-inline` helper.
- **An accuracy contract** — the silent-wrong-number traps: sync `font`/`lineHeight`/`letterSpacing` with CSS, avoid `system-ui`, await font load, clamp empty strings.
- Supported vs. out-of-scope CSS surface, common recipes (shrink-wrap, virtualization, flow-around-obstacle, inline chips), and a best-practices checklist.

It explicitly **points back to `README.md` as the authoritative API reference** (per the repo's `CLAUDE.md` convention), so it stays a usage layer rather than a second copy of the signatures that could drift.

## Notes

- API names in the skill were verified against the current `0.0.7` exports in `src/layout.ts` and `src/rich-inline.ts`.
- Scoped to consumer usage, not the contributor workflow (which `CLAUDE.md` / `DEVELOPMENT.md` already cover).
- Fully optional for anyone not using Claude Code — it's an inert Markdown file otherwise. Happy to relocate it (e.g. under `docs/`) or drop it if a tool-specific directory isn't wanted in the repo.

### DIFF
--- a/.claude/skills/pretext/SKILL.md
+++ b/.claude/skills/pretext/SKILL.md
@@ -1,0 +1,240 @@
+---
+name: pretext
+description: >-
+  How to use @chenglou/pretext for DOM-free multiline text measurement and line
+  layout. Use when measuring paragraph height without touching the DOM (avoiding
+  getBoundingClientRect/offsetHeight reflow), wrapping or laying out text to
+  canvas/SVG/WebGL/server-side, computing line counts or a multiline
+  shrink-wrap width, virtualizing long text lists, flowing text around
+  obstacles, or laying out inline rich text with chips/mentions. Triggers on
+  `prepare`/`layout`/`prepareWithSegments`/`layoutWithLines`/`walkLineRanges`,
+  the `@chenglou/pretext` or `@chenglou/pretext/rich-inline` imports, or any
+  "measure text height in JS without the DOM" task.
+---
+
+# Pretext
+
+Pure JS/TS library for **multiline text measurement and layout** with no DOM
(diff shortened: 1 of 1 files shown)


GuyNachshon:
also created this as a separate repo -> https://github.com/GuyNachshon/pretext-claude-skill

chenglou:
Thanks for this! It seems Claude has already picked up Pretext in its training data, so a skill might or might not be needed anymore. Thanks for hosting your own, though! Closing this one.

PR #208, opened 2026-07-07
fix: break overlong symbol runs

### LINKED ISSUE #206: "overflow-wrap: break-word;" not working as expected
Hey there, I am trying to use this library inside of a chat application, and noticed that the library doesn't calculate the right linebreaks for various repeated non-character symbols, such as ",{]\|\" and co:

https://codepen.io/editor/Stuben/pen/019f1dd9-d3f9-76d7-84e2-39093adb582e

In the example I made the pink elements be handled by browser, the blue via `prepare` and the pink one via `prepareRichInline`.

As per the projects readme I would expect this use case to work, and have both pretext calculated divs be the same size as the pink ones.

> `overflow-wrap: break-word`. Very narrow widths can still break inside words, but only at grapheme boundaries.

Image

hiSandog:
Summary:
- Allow multi-grapheme text segments that are not word-like to get overflow break advances.
- Add regression coverage for repeated symbol runs wrapping at grapheme boundaries.

Validation:
- /Users/sandog/.real/.bin/bun test src/layout.test.ts
- /Users/sandog/.real/.bin/bun run check

Fixes #206

### DIFF
--- a/src/layout.test.ts
+++ b/src/layout.test.ts
@@ -662,6 +662,15 @@ describe('prepare invariants', () => {
     expect(prepared.segments).toEqual([text])
   })
 
+  test('breaks overlong repeated symbol runs at grapheme boundaries', () => {
+    const prepared = prepareWithSegments('||||', FONT)
+    const maxWidth = measureWidth('|', FONT) + 0.1
+    const lines = layoutWithLines(prepared, maxWidth, LINE_HEIGHT).lines
+
+    expect(lines.map(line => line.text)).toEqual(['|', '|', '|', '|'])
+    expect(lines.every(line => line.width <= maxWidth)).toBe(true)
+  })
+
   test('keeps repeated punctuation runs attachable to trailing closing punctuation', () => {
     const prepared = prepareWithSegments('((()', FONT)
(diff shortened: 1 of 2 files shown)


Stubenhocker1399:
Still seems to overflow incorrectly for me using
```
{{{{{{{{{{{{{]]]]]]]]]]]]]\\\\\\\\\\\\\||||||||||||||||||||\{{{{{{{{{{{{{]]]]]]]]]]]]]\\\\\\\\\\\\\||||||||||||||||||||\{{{{{{{{{{{{{]]]]]]]]]]]]]\\\\\\\\\\\\\||||||||||||||||||||\{{{{{{{{{{{{{]]]]]]]]]]]]]\\\\\\\\\\\\\||||||||||||||||||||\{{{{{{{{{{{{{]]]]]]]]]]]]]\\\\\\\\\\\\\||||||||||||||||||||\
```
as the input text.

chenglou:
Thanks for digging into this! The repeated-symbol case and the mixed-symbol follow-up are fixed on main in #220. Your proposal helped shape the fix, and I added coauthor credit.

PR #211, opened 2026-07-21
fix: preserve leading ZWSP so zero-width-only lines are not dropped

### LINKED ISSUE #210: Bug: leading ZWSP (U+200B) is consumed at line start, one line undercounted vs. browsers
## Description

In CSS, U+200B (ZERO WIDTH SPACE) is **not** collapsible whitespace: it stays as
zero-width content at the start of a line and offers a break opportunity after it.
When a line starts with a ZWSP and the following word is too wide for the line box,
browsers wrap after the ZWSP and produce an empty first line containing only the ZWSP.

pretext's `consumesAtLineStart` treats `zero-width-break` like a space and skips it
at line start (via `normalizeLineStartSegmentIndex`), so it produces one line fewer
than the browser renders. This affects both the streaming and batch APIs.

## Reproduction

Open this [index.html](https://github.com/user-attachments/files/30215855/index.html)(recommended) or

Run in any browser page:

```js
const { prepareWithSegments, layoutWithLines } = await import(
  'https://unpkg.com/@chenglou/pretext@0.0.8/dist/layout.js'
)

const text = '\u200B≤100nA\u200B' // leading + trailing ZWSP
const prepared = prepareWithSegments(text, '12px Calibri, sans-serif', { whiteSpace: 'pre-wrap' })
const { lines } = layoutWithLines(prepared, 25, 20.96)
console.log(lines.map(l => JSON.stringify(l.text)))
```

DOM reference under the same conditions:

```html
<div style="width:25px; font:12px Calibri,sans-serif; line-height:20.96px;
            white-space:pre-wrap; overflow-wrap:break-word">&#8203;≤100nA&#8203;</div>
```

## Expected vs actual

| | lines |
|---|---|
| Browser DOM (verified in Chrome) | **3**: `"\u200B"` / `"≤100"` / `"nA\u200B"` |
| pretext 0.0.8 | **2**: `"≤100"` / `"nA\u200B"` |

Related edge case: a text of only `"\u200B"` yields **0** lines in pretext vs. **1** in browsers.

A self-contained repro page is attached (`index.html`): it loads pretext 0.0.8 from
unpkg, compares pretext output against DOM line counts across 5 cases, and shows the
`white-space` × `overflow-wrap` matrix.

## Root cause

`consumesAtLineStart` in `line-break.ts`:

```ts
function consumesAtLineStart(kind: SegmentBreakKind): boolean {
  return kind === 'space' || kind === 'zero-width-break' || kind === 'soft-hyphen'
}
```

ZWSP is not collapsible whitespace in CSS and should not be consumed at line start.
Note that simply removing `'zero-width-break'` here is not sufficient on the rich path:
a ZWSP-only line has `lineWidth === 0`, and `stepRichInlineLine` in `rich-inline.ts`
(`if (lineWidth === 0) return null`) would then drop that line and let the caller
terminate layout early — the "no progress" check needs to track fragment consumption
rather than line width.

ChuanfengZhang:
fix #210.
ZWSP was folded at line start like CSS-collapsible whitespace, so a line holding only the ZWSP was dropped and layout under-counted it versus real rendering. Keep ZWSP at line start and track fragment consumption separately from line width in the rich inline walker.

### DIFF
--- a/src/line-break.ts
+++ b/src/line-break.ts
@@ -34,7 +34,12 @@ type InternalLineVisitor = (
 ) => void
 
 function consumesAtLineStart(kind: SegmentBreakKind): boolean {
-  return kind === 'space' || kind === 'zero-width-break' || kind === 'soft-hyphen'
+  // ZWSP is not CSS-collapsible whitespace: it is zero-width content that
+  // survives on the line and offers a break opportunity after it. When the
+  // next word overflows, the browser breaks after the ZWSP, producing a line
+  // that holds only the ZWSP. Consuming it at line start would under-count
+  // that line versus real rendering.
+  return kind === 'space' || kind === 'soft-hyphen'
 }
 
 function breaksAfter(kind: SegmentBreakKind): boolean {
--- a/src/rich-inline.ts
+++ b/src/rich-inline.ts
@@ -232,6 +232,12 @@ function stepRichInlineLine(
   let lineWidth = 0
   let remainingWidth = safeWidth
   let itemIndex = cursor.itemIndex
+  // Zero-width fragments (e.g. a leading ZWSP alone on a line) do not add to
+  // lineWidth but still consume the cursor and form a real line. Track via a
+  // consumed flag rather than lineWidth === 0, otherwise the zero-width line
+  // is dropped and the caller terminates layout early because the cursor
+  // never advanced.
+  let consumedAnyFragment = false
 
   lineLoop:
   while (itemIndex < flow.items.length) {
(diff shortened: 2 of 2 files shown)


chenglou:
Thanks for the fix and repro! The rich-inline part is now addressed in #220, with coauthor credit. Keeping this open while I finish the flat leading-ZWSP case.

chenglou:
Thanks again for this one, and for the repro in #210! The rich-inline half went in with #220 a while back, and the flat fix just landed in #227. It keeps a leading ZWSP at the start of a paragraph or after a hard break. It still drops one right after a wrap or a forced break, since browsers do some odd things there that Pretext doesn't model yet. Your example is now a required check in Chrome, Safari and Firefox, so it won't quietly regress. Closing this one as superseded.

PR #213, opened 2026-07-21
fix: avoid breaking after a numeric minus sign

### LINKED ISSUE #212: Bug: numeric minus sign is treated as a preferred break point (`-0.475`, `≥-100nA`)
## Description

pretext treats ASCII `-` as a preferred soft-break grapheme. That is correct for
compound words such as `well-known`, but incorrect when `-` is a **numeric minus**
that belongs to the following number — either at the start of a run (`-0.475`) or
after a non-alphanumeric character such as a comparison symbol (`≥-100nA`).

When the available width forces an emergency (mid-segment) break, pretext prefers to
break **after** that minus sign. Browsers keep the sign with the number and only break
at a later grapheme boundary. Depending on the case, this produces an extra short line
(`"-"` on its own line) or a split that separates the sign from its number
(`"≥-"` | `"100nA"`).

## Reproduction

Run in any browser page:

```js
const { prepareWithSegments, layoutWithLines } = await import(
  'https://unpkg.com/@chenglou/pretext@0.0.8/dist/layout.js'
)

const font = '16px Arial, sans-serif'
const ctx = document.createElement('canvas').getContext('2d')
ctx.font = font
const widthOf = (t) => ctx.measureText(t).width

// Case 1: leading numeric minus — width fits "-0.47" but not the whole token
const p1 = prepareWithSegments('-0.475', font)
console.log(layoutWithLines(p1, widthOf('-0.47') + 0.1, 20).lines.map(l => l.text))
// pretext 0.0.8: ["-", "0.47", "5"]   ← breaks after the leading "-"
// browsers:      ["-0.47", "5"]

// Case 2: numeric minus after a comparison symbol
const p2 = prepareWithSegments('≥-100nA', font)
console.log(layoutWithLines(p2, widthOf('≥-100n') + 0.1, 20).lines.map(l => l.text))
// pretext 0.0.8: ["≥-", "100nA"]      ← sign split from its number
// browsers:      ["≥-100n", "A"]
```

DOM reference under the same conditions:

```html
<div style="width:36.57px; font:16px Arial,sans-serif; line-height:20px;
            overflow-wrap:break-word">-0.475</div>
```

## Expected vs actual

(Widths are font-dependent; values below are for 16px Arial in Chrome.)

| case | | lines |
|---|---|---|
| `-0.475` @ ~36.6px | Browser DOM | **2**: `"-0.47"` / `"5"` |
| | pretext 0.0.8 | **3**: `"-"` / `"0.47"` / `"5"` |
| `≥-100nA` @ ~49.8px | Browser DOM | **2**: `"≥-100n"` / `"A"` |
| | pretext 0.0.8 | **2**: `"≥-"` / `"100nA"` |
| control: `well-known` | both | `"well-"` / … (hyphen still preferred) |
| control: `x-100` | both | `"x-"` / `"100"` (`-` after a letter is an ordinary hyphen) |

A self-contained repro page is attached ([index.html](https://github.com/user-attachments/files/30221886/index.html)): it loads pretext 0.0.8 from
unpkg, prints the per-line splits of both pretext and the DOM side by side, and
includes the two hyphen controls above.

## Root cause

In `layout.ts`, `getBreakablePreferredBreaks` marks every preferred-break grapheme
(including ASCII `-`) without checking whether it is a numeric sign:

```ts
for (const gs of getSharedGraphemeSegmenter().segment(text)) {
  graphemeIndex++
  if (isPreferredBreakGrapheme(gs.segment)) breaks.push(graphemeIndex)
}
```

A rule that matches browser behavior for the cases above:

- `-` is a numeric sign when it is followed by a decimal digit (`\p{Nd}`), and
- the preceding grapheme is either absent or not a letter/number (`[\p{L}\p{N}]`);
- such a `-` should not be a preferred break. Ordinary hyphens (`well-known`,
  `x-100`) stay preferred, and the emergency grapheme-level path can still split
  later if the whole token does not fit.

ChuanfengZhang:
fix #212.
A numeric minus sign—whether at the leading position (e.g. `-0.475`) or following a non-alphanumeric character such as a comparison symbol (e.g. `≥-100nA`)—is no longer treated as a preferred break point. The sign stays with its number and the break falls to a later grapheme boundary, matching browser line-breaking behavior.

### DIFF
--- a/src/layout.test.ts
+++ b/src/layout.test.ts
@@ -532,6 +532,34 @@ describe('prepare invariants', () => {
     expect(prepareWithSegments('-1/12', FONT).segments).toEqual(['-1/12'])
   })
 
+  test('does not prefer a leading numeric sign over a later emergency break', () => {
+    const text = '-0.475'
+    const prepared = prepareWithSegments(text, FONT)
+    const width = measureWidth('-0.47', FONT) + 0.1
+
+    expect(prepared.segments).toEqual([text])
+    expect(prepared.breakablePreferredBreaks).toEqual([null])
+
+    const batched = layoutWithLines(prepared, width, LINE_HEIGHT)
+    expect(batched.lines.map(line => line.text)).toEqual(['-0.47', '5'])
+    expect(collectStreamedLines(prepared, width)).toEqual(batched.lines)
+    expect(layout(prepared, width, LINE_HEIGHT).lineCount).toBe(2)
+    expect(measureLineStats(prepared, width).lineCount).toBe(2)
+  })
+
+  test('does not prefer a numeric sign after a comparison symbol', () => {
+    const text = '≥-100nA'
+    const prepared = prepareWithSegments(text, FONT)
+    const width = measureWidth('≥-100n', FONT) + 0.1
+    const lines = layoutWithLines(prepared, width, LINE_HEIGHT).lines.map(line => line.text)
+
+    expect(prepared.segments).toEqual([text])
+    expect(prepared.breakablePreferredBreaks).toEqual([null])
+    expect(lines).toHaveLength(2)
+    expect(lines.join('')).toBe(text)
+    expect(lines[0]).not.toBe('≥-')
+  })
+
   test('keeps URL-like runs together as one breakable segment', () => {
     const prepared = prepareWithSegments('see https://example.com/reports/q3?lang=ar&mode=full now', FONT)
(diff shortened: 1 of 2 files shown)


chenglou:
Thanks for the fix and the repros! I folded this into the wrapping cleanup in #220, with coauthor credit. It’s on main now, so closing this as superseded.

PR #215, opened 2026-07-22
fix: align CJK ASCII hyphen wrapping with browsers

### LINKED ISSUE #214: BUG: CJK text can wrap with an ASCII hyphen at the start of a line
## Description

Pretext can choose different line breaks from the browser when an ASCII hyphen (`U+002D`) appears between CJK text.

For the following text:

```text
(试验前-试验后)/试验前
```

and a width just large enough for `前-试验`, the browser keeps the hyphen with the preceding CJK character. Pretext instead allows the hyphen to begin the
next line.

This reproduces with `@chenglou/pretext@0.0.8` and is also present on current `upstream/main` at `ac49b09`.

## Reproduction

```js
import {
  layoutWithLines,
  prepareWithSegments,
} from '@chenglou/pretext'

const text = '(试验前-试验后)/试验前'
const font = '20px Arial'
const context = document.createElement('canvas').getContext('2d')
context.font = font

const width = context.measureText('前-试验').width + 0.1
const prepared = prepareWithSegments(text, font, { whiteSpace: 'pre-wrap' })
const lines = layoutWithLines(prepared, width, 28).lines.map(line => line.text)

console.log(prepared.segments)
console.log(lines)
```

A standalone HTML reproduction is also available [pretext-cjk-ascii-hyphen-wrapping-repro.html](https://github.com/user-attachments/files/30259405/pretext-cjk-ascii-hyphen-wrapping-repro.html).

It renders the same text and width through both native DOM layout and Pretext, then prints both line arrays and the prepared segments.

Verified environment:

- macOS
- Google Chrome 150.0.7871.129
- `@chenglou/pretext@0.0.8`

## Expected vs actual

Browser DOM:

```js
['(试验', '前-试验', '后)/试', '验前']
```

Pretext 0.0.8:

```js
['(试验前', '-试验', '后)/试', '验前']
```

The corresponding Pretext segments include separate `前`, `-`, and `试` units, which makes the line-start hyphen possible.

## Root cause

The CJK kinsoku line-start set does not include ASCII `-`. As a result,
`buildBaseCjkUnits()` does not attach an ASCII hyphen to the preceding CJK
unit, even though browser layout keeps the pair together in this context.

Adding `-` to the line-start-prohibited set alone is not sufficient:

1. In text such as `温度-5度`, the hyphen is a numeric sign. Merging it only
   backward can separate it from the following digits and introduce a new
   preferred break after the sign.
2. A merged CJK-hyphen or CJK-numeric unit can be wider than an empty line. It
   still needs the normal grapheme-level emergency fallback used for
   `overflow-wrap: break-word` behavior.

## Proposed solution

- Treat ASCII `-` as line-start-prohibited in CJK segmentation, attaching it to
  the preceding CJK unit.
- When that CJK-adjacent hyphen is followed by a decimal digit, keep the numeric
  suffix together and suppress a preferred break after the sign.
- Mark CJK-hyphen units as eligible for grapheme-level emergency breaking when
  the complete unit cannot fit on an empty line.
- Preserve that emergency-break capability when units are merged for
  `wordBreak: 'keep-all'`.
- Keep ordinary non-CJK behavior unchanged, including the existing preferred
  break in `foo-5`.

The numeric-sign guard should be scoped to `CJK + "-" + decimal digit`. This
fix intentionally does not change leading or symbol-prefixed numeric signs such
as `-0.475` or `≥-100nA`, which are a separate issue.

Suggested regression coverage:

- `(试验前-试验后)/试验前` matches browser line wrapping.
- `温度-5度` and `気温-10度です` keep the sign with both the CJK prefix and
  numeric suffix.
- An overlong `温度-100nA` unit can still emergency-break at grapheme
  boundaries.
- `日本語foo-5` retains the ordinary Latin hyphen break behavior.
- `layout()`, `layoutWithLines()`, `layoutNextLine()`, and line-stat helpers stay
  aligned.

ChuanfengZhang:
fix #214


pyrex41:
Nice fix — the numeric-minus handling (gluing `度-5` on both sides, gated on `unitContainsCJK`, with `allowOverflowBreaks` as the escape hatch for overlong glued units) reads carefully, and the `日本語foo-5` test is a good discriminator showing Latin `foo-` still breaks.

One clarifying question before it lands: the numeric-minus logic is CJK-gated, but the `'-'` addition to `kinsokuStart` is **global** — it now prohibits a hyphen from starting *any* wrapped line, not just CJK-adjacent ones. Have you confirmed that doesn't change line-start behavior in non-CJK contexts where a wrapped line would legitimately begin with a hyphen-minus? I'm thinking of things like code/identifier text or numeric ranges (`a - b`, `foo -bar`, an option list) where the browser *would* allow the `-` at the start of a line. The existing Latin/URL tests suggest it's fine, but a test that pins the non-CJK "hyphen may start a line" case (asserting no regression vs. `main`) would make the global `kinsokuStart` change safe to reason about.

Not a blocker — just want to be sure the global set change is intentional and side-effect-free outside CJK.

pyrex41:
Followed up on my own question — I traced it and diffed both branches, and the answer is reassuring: **the global `kinsokuStart` `'-'` addition is inert outside CJK.** Sharing so you don't have to redo the legwork.

**Trace** — all three consult sites are CJK-guarded or already shadowed:
- `buildBaseCjkUnits` (layout.ts) is only called under `if (segKind === 'text' && segMetrics.containsCJK)`, so pure-Latin segments never reach it. For Latin sub-runs *inside* a mixed segment the new `kinsokuStart.has(grapheme)` branch calls the same `appendToUnit` the old fall-through did — it only diverges when `unitContainsCJK` is already true (the bug being fixed).
- `isCJKLineStartProhibitedSegment` (analysis.ts) is consumed only in the `&& tailContainsCJK` branch of `buildMergedSegmentation`.
- The keep-all path already had `-` in `keepAllDashBreakChars`, so `canContinueKeepAllTextRun` returns the same result before/after.

**Empirical** — built both branches (bun runs the TS sources directly with the repo's fake-canvas backend) and diffed `prepareWithSegments` + `layoutWithLines`:
- 33 adversarial non-CJK inputs (`a - b`, `foo -bar -baz`, `--flag -x`, `x = a-b`, `well-known`, option lists…), each width-swept 3px→natural in 0.5px steps under default / pre-wrap / keep-all / break-all / break-word → **byte-identical** main vs PR.
- A 400-string fuzz including Latin, digits, Arabic, Hebrew, Devanagari, Thai → ~19k lines/side, **identical**.
- Positive control: the #214 CJK repro *does* change (44 differing lines across the sweep), so the harness isn't blind — the non-CJK identity is real, not a false negative.
- `bun test` on the branch: 120 pass, 0 fail.

So this is safe as-is. The only optional polish would be gating the set membership behind a CJK check (or adding a non-CJK no-regression test) to make the intent legible — purely cosmetic, not a correctness issue. Nice fix.

ChuanfengZhang:
Thank you @pyrex41  so much for digging into this so thoroughly, and especially for following up on your own question with the trace, branch diff, width sweeps, fuzzing, and positive control. That saved me a ton of legwork and gave me much more confidence that the global kinsokuStart addition is truly inert outside CJK.
I really appreciate how patient and concrete this review was. I’ve pushed a small non-CJK regression test using foo -bar, just to make the intent explicit and guard against future call sites accidentally broadening the behavior. Honestly, this was an exceptionally thoughtful review. Thanks again!

chenglou:
Thanks for the fix and for following up on the CJK/numeric edge cases! I folded the useful bits into #220, with coauthor credit. The extra Latin-hyphen control is covered too; closing this as superseded.

PR #217, opened 2026-08-24
perf: minor optimizations in benchmark and build scripts

### LINKED ISSUE #179: Minor Optimisations
I would have done a pull request had I been bothered to do a fork, but here are my things:
all instances of `math.floor(n/2)`, such as the one in scripts/benchmark_check.ts should become `n>>1`, mathematically identical but faster as JS can not handle any iterable longer than 2^32-1 words, and the line,
`return sorted.length % 2 === 0 ? (sorted[mid - 1]! + sorted[mid]!) / 2 : sorted[mid]! `
could become
`return sorted.length & 1 ? sorted[mid]! : (sorted[mid - 1]! + sorted[mid]!) / 2 `
which is equivilant but probably slightly faster, and if you don't like the re-ordering, you can do
`return ~(sorted.length & 1) ? (sorted[mid - 1]! + sorted[mid]!) / 2 : sorted[mid]! `

and in scripts/build-demo-site.ts
```
if (result.exitCode !== 0) {
  process.exit(result.exitCode)
}
```
could become
```
if (result.exitCode) {
  process.exit(result.exitCode)
}
```
, which I think is neater, with no performance disadvantages that I know of, and this can be generalised to replace all `if(num!=0)` with `if(num)`

Also, on line 330, `process.exitCode = 1` maybe could become `process.exitcode = 1&0xFF` to make it obvious to the browser that this is not a float but an integer that fits in 8 bits, but take this with a grain of salt as this does feel a bit pointless and contradicts my previous point about neatness.

The same goes for adding `|0` after integer declarations and simple reassignments e.g `x=y` to make it obvious to a bad compiler that it's an integer to maybe speed it up, but again, this feels pointless and probably is, alongside being not particularly elegant so I would not recommend it, but you can consider it.

3ni8ma:
Implements minor optimizations suggested in issue #179.

Changes:
- Use bitwise shift instead of Math.floor division in median()
- Use bitwise AND for parity check instead of modulo
- Simplify exitCode check to truthy test

Closes #179

### DIFF
--- a/scripts/benchmark-check.ts
+++ b/scripts/benchmark-check.ts
@@ -96,8 +96,8 @@ function parseBrowser(value: string | null): BrowserKind {
 
 function median(values: number[]): number {
   const sorted = [...values].sort((a, b) => a - b)
-  const mid = Math.floor(sorted.length / 2)
-  return sorted.length % 2 === 0 ? (sorted[mid - 1]! + sorted[mid]!) / 2 : sorted[mid]!
+  const mid = sorted.length >> 1
+  return sorted.length & 1 ? sorted[mid]! : (sorted[mid - 1]! + sorted[mid]!) / 2
 }
 
 function assertSame<T>(actual: T, expected: T, context: string): void {
--- a/scripts/build-demo-site.ts
+++ b/scripts/build-demo-site.ts
@@ -25,7 +25,7 @@ const result = Bun.spawnSync(
   },
 )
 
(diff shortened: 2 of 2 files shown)


chenglou:
Thanks! But see my comment at https://github.com/chenglou/pretext/issues/179#issuecomment-5627219768. It feels these have very minor gains and I'm not sure they're worth it, but please do keep the fixes coming if they affect speed measurably!

PR #218, opened 2026-09-03
fix: break overlong symbol runs at grapheme boundaries

### LINKED ISSUE #206: "overflow-wrap: break-word;" not working as expected

Dahhrk:
Overlong non-word text segments (repeated symbols like `|`, `{`, `]`, `\\`) were not getting grapheme-level overflow breaks because `pushMeasuredTextSegment` gated breakable fit advances on the `wordLike` flag. Symbol runs are not word-like, so they overflowed the container.

## Changes

- Guard is `kind === 'text'` instead of `wordLike`, so any text segment gets grapheme-level overflow breaks when it exceeds available width.
- Drop the unused `wordLike` parameter from `pushMeasuredTextSegment`.
- Regression test: simple `||||` plus mixed `{{{]]]\\|||`.

## Why PR #208 was closed

#208 applied the same `kind === 'text'` fix and tested only `||||`. A longer mixed symbol run still overflowed. This reuses that approach and adds the mixed-symbol case.

Fixes #206

### DIFF
--- a/src/layout.test.ts
+++ b/src/layout.test.ts
@@ -662,6 +662,22 @@ describe('prepare invariants', () => {
     expect(prepared.segments).toEqual([text])
   })
 
+  test('breaks overlong symbol runs at grapheme boundaries', () => {
+    const pipeWidth = measureWidth('|', FONT)
+    const maxWidth = pipeWidth + 0.1
+
+    const simplePrepared = prepareWithSegments('||||', FONT)
+    const simpleLines = layoutWithLines(simplePrepared, maxWidth, LINE_HEIGHT).lines
+    expect(simpleLines.map(l => l.text)).toEqual(['|', '|', '|', '|'])
+    expect(simpleLines.every(l => l.width <= maxWidth)).toBe(true)
+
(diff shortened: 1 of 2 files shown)


chenglou:
Thanks for picking this up! #220 fixed #206 in the meantime, mixed runs like `{{{]]]\\|||` included, and emoji runs keep their current behavior. So this one's superseded, but the mixed-symbol test was a nice catch.

Issue #202, opened 2026-06-24
Markdown Chat example renders in Firefox with text overflowing speech bubbles


brandonzylstra:
This is Firefox version 151.0.1 (aarch64) on macOS 15.6.1 (24G90) Sequoia.

<img width="1165" height="530" alt="Image" src="https://github.com/user-attachments/assets/70532759-5a08-4f98-814f-fd578b282441" />

This is the only problem I've noticed in Firefox.   [Bubbles](https://chenglou.me/pretext/bubbles/) and all other demos seem work beautifully.

hiSandog:
Helpful report. Since the other demos work in Firefox, this may be a Markdown Chat example/style interaction rather than a core layout failure. It would be useful to add the exact font/zoom settings and whether the overflowing bubble contains a long unbreakable token or normal wrapped text, because Firefox can differ from Chromium around min-content sizing and `overflow-wrap`/`white-space` combinations.

brandonzylstra:
@hiSandog I have the font zoomed at 100%, but even if I go above or below 100% I still see the same effect.

This might be slightly helpful:

<img width="1313" height="625" alt="Image" src="https://github.com/user-attachments/assets/cf2953d0-840f-42b9-855e-db3694dc1aab" />

chenglou:
Thanks for the report and screenshots! Fixed in [871e48f](https://github.com/chenglou/pretext/commit/871e48f7ca6a4a260508ec5d7f9e58ab19ab4e65). Firefox was resolving the system font differently for canvas measurement and DOM text; using a named font stack fixes the overflowing bubbles. Checked in Firefox and Chrome on macOS.

PR #219, opened 2026-09-03
docs+test: break:'never' chips stay atomic; demo pre-wrap is not a chip

### LINKED ISSUE #203: Chips not "staying whole" in Firefox
It's unclear to me whether this is a bug or whether I'm misunderstanding the intended meaning of a chip "staying whole".  The "pre-wrap" chip is splitting into two chips when the text-width is set to just the right value.  That _appears (to me)_ to contradict the description of chips staying whole.  However, perhaps the intention was that the chips maintain visual integrity as whole chips, rather than appearing to be two pieces of a single chip when the text breaks.  In that case, the chips are performing as intended.

I could envision an argument being made for either behavior, or for the ability to specify the desired behavior.

Image

If I've misunderstood the description, please consider clarifying the wording.  Thanks!

Dahhrk:
Fixes https://github.com/chenglou/pretext/issues/203

The `break: 'never'` engine already keeps a chip on one line as one fragment. The demo "pre-wrap chip" is a code span with `break: 'normal'`. It can split. That is not a chip.

This PR does not change layout engine code.

- README + API comments: `break: 'never'` stays whole (wraps as a unit or overflows). `extraWidth` is chrome only.
- Three tests in `src/layout.test.ts`: atomic across widths, overflow stays one fragment, wrap-to-next-line as a whole.

Not in this PR: #202, #210, #212, #214.

Agent claimed `bun test` 120 pass and `bun run check` clean. I did not re-run those here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ walkRichInlineLineRanges(prepared, 320, range => {
 It is intentionally narrow:
 - raw inline text in, including boundary spaces
 - caller-owned `extraWidth` for pill chrome
-- `break: 'never'` for atomic items like chips and mentions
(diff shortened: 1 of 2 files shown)


chenglou:
Hey! Thanks, the gray ones are code blocks, not chips. The chips don't break. I've clarified this by making the code blocks' border radius smaller

